PR #2, opened 2024-05-20
Open EMME GUI after creating EMME project

zptro:
If enabling `visible`, `_app.start()` will be used instead of `_app_start_dedicated()`, so that it will stay active after python script ends.


juliuskroetzl:
I tested running create_emme_project.py, but at the end of the script, the EMME GUI closed by itself. Was it intended to remain open so that the user could for example, import a scenario?

juliuskroetzl (approved):
Branch works fine when using VLEM UI

PR #3, opened 2024-05-20
Migrate from os.path to pathlib

zptro:
Also change the path for result matrices from `submodel_name/Matrices` to `Matrices/submodel_name` and for base matrices from `base_matrices/submodel_name` to `Matrices/submodel_name`. This makes result matrix use as base matrices more straightforward.


Scripts/lem.py
```diff
@@ -39,15 +41,15 @@ def main(args):
         }
     }
     # Check input data folders/files exist
-    if not os.path.exists(base_zonedata_path):
+    if not base_zonedata_path.exists():
```
samakinen, line 44: These checks could be is_dir() instead of exists() because we expect the paths to be directories later. Not strictly required, but it might help to catch some incorrect arguments.

Scripts/modelsystem.py
```diff
@@ -47,37 +46,31 @@ class ModelSystem:
...
     """
 
-    def __init__(self, 
-                 zone_data_path: str, 
-                 base_zone_data_path: str, 
-                 base_matrices_path: str,
-                 results_path: str, 
-                 assignment_model: AssignmentModel, 
-                 name: str,
+    def __init__(self,
```
samakinen, line 53: The class docstring will still have type annotations for str instead of Path

samakinen (approved):
Looks really good and makes the code more concise. 
I added couple minor comments regarding Path.is_dir() vs Path.exists() when we expect to find directories instead of files. Also some docstrings still have the old str types.

PR #4, opened 2024-05-21
Add reading and aggregation of long-distance trips in submodels

zptro:
The main feature is the new method `ModelSystem.add_external_demand()` that is activated for short trip (i.e., submodel) model runs. This method first looks for disaggregated ("koko_suomi") long-distance trip matrices in the base matrix folder and tries to aggregate them. The disaggregated format implies that the long-distance model has been run recently.

If the disaggregated matrices are not found, the next step is to look for aggregated long-distance matrices. These aggregated base matrices should be available for all the submodels, in case the user is not interested in long-distance demand changes. They must fit the zone system of the submodel in question.

- This change has some indirect effects on handling of aggregations, which is from now on done both for zone data and for matrices.
- `--end-assignment-only` model runs will assign matrices from the base matrix folder instead of results folder.
- Added copying of the input data to result folder, so that the result folder can be used as base input data in another model run.


Scripts/modelsystem.py
```diff
@@ -64,6 +65,10 @@ def __init__(self,
         self.zdata_base = BaseZoneData(
             base_zone_data_path, self.zone_numbers, f"{submodel}.zmp")
         self.basematrices = MatrixData(base_matrices_path / submodel)
+        self.mapping = pandas.read_csv(
```
zptro, line 68: This should be done in only one place, so I should change `read_csv`.

attesn:
Can you add brief description and context for pull request? I'm a bit unsure what I'm reviewing.

Scripts/modelsystem.py
```diff
@@ -271,14 +305,8 @@ def run_iteration(self, previous_iter_impedance, iteration=None):
...
 
-        # Calculate internal demand
         self._add_internal_demand(previous_iter_impedance, iteration=="last")
-
-        # Calculate external demand
-        for mode in param.external_modes:
-            int_demand = self._sum_trips_per_zone(mode)
-            ext_demand = self.em.calc_external(mode, int_demand)
-            self.dtm.add_demand(ext_demand)
+        self._add_external_demand()
```
zptro, line 309: At this stage, only car matrices should be added, because the other matrices are not initialized.

Scripts/modelsystem.py
```diff
@@ -163,6 +167,50 @@ def _add_internal_demand(self, previous_iter_impedance, is_last_iteration):
...
+                # Try getting long-distance trips from separate file
+                with self.long_dist_matrices.open(
+                        "demand", "vrk", zone_numbers,
+                        self.mapping, long_dist_classes) as mtx:
+                    for ass_class in long_dist_classes:
+                        demand = Demand(self.em.purpose, ass_class, mtx[ass_class])
+                        self.dtm.add_demand(demand)
+                        if ass_class in param.car_classes:
+                            car_matrices[ass_class] = demand.matrix
+            except IOError:
```
attesn, line 189: I'd find it better to have user to do this choice when starting model run. I think this way user might not know what assumptions her/his model entails. Also model run might be erroneous without user ever knowing it.

Other option would be to ask user to point out long distance matrices and interrupt if not found. As minimal change solution I would propose `log.warning` message here.

attesn (commented):
Good work! This concept of long-distance trip aggregation seems good, but I had some doubts with not-directly-expressed selection of long distance matrices. Detailed comments below.

Scripts/modelsystem.py
zptro: Yes, you are right. I now added logging of the events here, but in the next sprint I will make changes to the run-time parameters (we should also change the name of the `--free-flow-assignment` flag to `--long-distance-demand` or similar) and have Maaru do the corresponding updates to the UI.

Scripts/datahandling/zonedata.py
```diff
@@ -223,31 +223,28 @@ def get_data(self, key: str, bounds: slice, generation: bool=False) -> Union[pan
...
             peripheral[1])]
-        municipality_file = "koko_suomi_kunta.zmp"
-        municipality_centre_zones = pandas.read_csv(
-            data_dir / municipality_file, delim_whitespace=True,
-            index_col="data_id").squeeze().drop_duplicates().sort_values()
-        files = FileReader(
-            data_dir, municipality_centre_zones,
-            zone_mapping_file=municipality_file)
-        mapping = files.read_csv_file(".agg")["municipality"]
+        municipality_centre_mapping = pandas.read_csv(
```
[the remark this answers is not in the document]
zptro: Good point. Fixed.

Scripts/lem.py
```diff
@@ -53,6 +54,13 @@ def main(args):
         raise NameError(
             "Forecast data directory '{}' does not exist.".format(
                 forecast_zonedata_path))
+    shutil.copytree(
```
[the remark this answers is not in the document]
zptro: I agree it is not a perfect fit. But it relates to the changed behaviour of `--end-assignment-only`. When both matrices and zone data are found in the results folder, that folder can be used as input data to another model run (e.g., end-assignment-only).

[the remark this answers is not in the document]
attesn: Ok! We can keep this as it is.

attesn (approved):

PR #6, opened 2024-05-22
Fix zonedata.py and zone.py to process new input data. 

EliasMarttunen:
New input data is separately from years 2016 and 2023 and it also resets trips from Åland to zero.


zptro (approved):
Looks good!

PR #7, opened 2024-05-28
Calibration 1

EliasMarttunen:
Mode choise constants calibrated.


attesn (commented):
Good work! Seems that calibration equation is working well. I think we can close this one and wait for 2016 estimation round, as it should be ready this week.

Next time Pull Request name could be more informative `Calibration of mode choice constants` (etc) as this remains visible in commit [logbook](https://github.com/Traficom/lem-model-system/commits/main/) and self-explanatory name enables having a quick overlook of changes made. Then again commits made within PR could have round tag.

attesn:
I'll close this as we'll continue calibration from 2016 estimated parameters.

PR #8, opened 2024-05-28
Use cost damping in mode-destination choice models

attesn:
Pull Request to test cost damping as a solution for underestimated tour lengths. Here cost damping is done by using linear and nonlinear ´cost_c + log(cost_c)´ factors in utility functions. 

Parking cost was moved to car cost as a practical solution, as many tour types would get positive cost coefficients with separate parking cost + linear and nonlinear cost parameters.

First edition done without any updates to distance boundaries. We should test also allowing >100 km distances and using overnight tours in place of long tours.


attesn:
@eliasaxel Updated by fixing errors in parking cost calculation. Could you run validation again?

PR #10, opened 2024-05-30
Update model result files

attesn:
Update model result files:
* Rename files to correspond the contents.
* Remove results that are not working correctly, caused by previous model-system changes. 
* Exports are collected to one modelsystem function.
* Result summary of mode shares are logged to be visible for user.


Scripts/modelsystem.py
```diff
@@ -373,49 +346,76 @@ def _calculate_noise_areas(self):
...
-            if modes == "sustainable":
-                intervals = zone_param.savu_intervals
-                savu = numpy.searchsorted(intervals, access[modes][bounds]) + 1
+            accessibility = purpose.accessibility_model.accessibility
+            for key in accessibility:
+                logsum = pandas.Series(numpy.log(accessibility[key]), 
+                    purpose.zone_numbers, name=f"{purpose.name}_{key}")
+                self.resultdata.print_data(logsum, f"accessibility.txt")
+    
+    def _export_resultdata(self):
```
zptro, line 357: I think it is a good idea to have all the result printing here. I think a good compromise (considering oop) and a nice solution for making this method cleaner would be to have several simple `@property` getter methods in `TourPurpose`. For instance:

```
@property
def all_generated_tours(self):
    return pandas.Series(
        sum(self.generated_tours.values()), self.zone_numbers, name=self.name)
```

The same goes for `modes_shares`, `trip_lengths`, etc.

Scripts/modelsystem.py
```diff
@@ -373,49 +346,76 @@ def _calculate_noise_areas(self):
...
-                int_demand[bounds] += purpose.generated_tours[mode]
-                if include_dests:
+                if generation:
+                    int_demand[bounds] += purpose.generated_tours[mode]
+                if attraction:
                     int_demand += purpose.attracted_tours[mode]
         return int_demand
+    
+    def _get_purpose_tours(self, purpose, generation = True, attraction = True):
+        int_demand = pandas.Series(0, self.zdata_base.zone_numbers, name=purpose.name)
```
zptro, line 413: This would not work with purpose bounds.

Scripts/modelsystem.py
```diff
@@ -156,11 +156,11 @@ def _add_internal_demand(self, previous_iter_impedance, is_last_iteration):
                         purpose, "car_leisure", purpose_impedance)
             else:
                 demand = purpose.calc_demand()
-                if is_last_iteration:
-                    purpose.print_data()
                 if purpose.dest != "source":
                     for mode in demand:
                         self.dtm.add_demand(demand[mode])
+        if is_last_iteration:
+            self._export_resultdata()
```
zptro, line 163: I think we should move this to the end of `run_iteration()`. Then it will automatically affect agent model as well.

Scripts/modelsystem.py
```diff
@@ -276,38 +272,10 @@ def run_iteration(self, previous_iter_impedance, iteration=None):
...
-        # Calculate tour sums and mode shares
-        tour_sum = {mode: self._sum_trips_per_zone(mode, include_dests=False)
-            for mode in self.travel_modes}
-        sum_all = sum(tour_sum.values())
-        mode_shares = {}
-        agg = self.zdata_base.aggregations
-        self.resultdata.print_data(tour_sum, "origins_demand.txt")
-        self.resultdata.print_data(
-            {mode: tour_sum[mode] / sum_all for mode in tour_sum},
-            "origins_shares.txt")
```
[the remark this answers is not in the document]
zptro: We discussed this, and I agree that it may as well be part of result post-processing.

Scripts/modelsystem.py
```diff
@@ -373,49 +346,76 @@ def _calculate_noise_areas(self):
...
-                avg_savu = numpy.searchsorted(intervals, avg_access) + 1
-                avg_savu += ((avg_access - intervals[avg_savu-2])
-                            / (intervals[avg_savu-1] - intervals[avg_savu-2]))
-                self.resultdata.print_line(
-                    f"Average SAVU:\t{avg_savu:1.4f}", "result_summary")
-
-    def _sum_trips_per_zone(self, mode, include_dests=True):
+                    purpose.within_zone_tours[mode], "within_zone_tours.txt")
+
+    def _get_mode_tours(self, mode, generation = True, attraction = True):
```
zptro, line 399: This is just a small remark, but if both are `True` (the default), does it not return the number of *trips* (departing+arriving)?

Scripts/models/logit.py
```diff
@@ -476,36 +476,21 @@ def calc_accessibility(self, impedance):
...
-            workforce = ((normalization*mode_expsum)
-                            **(1/next(iter(param.values()))["log"]["logsum"]))
-            workforce = pandas.Series(
-                workforce, self.purpose.zone_numbers, name=self.purpose.name)
-            self.resultdata.print_data(workforce, "workplace_accessibility.txt")
+                self.accessibility["sustainable"] += numpy.log(self.mode_exps[mode])
+        money_utility /= next(iter(self.mode_choice_param.values()))["log"]["logsum"]
+        self.accessibility["all_scaled"] = money_utility * self.accessibility["all"]
+        self.accessibility["sustainable_scaled"] = money_utility *  self.accessibility["sustainable"]
+        self.accessibility["car_scaled"] = money_utility * self.accessibility["car"]
```
zptro, line 493: Should we do a for-loop for these modifications?

zptro (changes requested):
Thank you for this effort! With some fine-tuning we will get a much nicer solution than the current one.

Scripts/modelsystem.py
```diff
@@ -328,13 +296,18 @@ def run_iteration(self, previous_iter_impedance, iteration=None):
                 self.resultdata,
                 self.zdata_base.aggregations.municipality_mapping)
             self._calculate_noise_areas()
-            self._calculate_accessibility_and_savu_zones()
-            self.resultdata.print_line("\nMode shares", "result_summary")
-            for mode in mode_shares:
-                self.resultdata.print_line(
-                    "{}\t{:1.2%}".format(mode, mode_shares[mode]),
-                    "result_summary")
```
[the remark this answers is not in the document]
attesn: I thought `results_summary.txt` to be just check if model run gives reasonable results and that it would be better to be visible in log. However, if result summary file is needed I can restore it.

Scripts/modelsystem.py
attesn: Yep, agreed. Didn't come with any good ideas of how implement this function (both generation and attraction). Probably having them separately as purpose methods is better.

Scripts/modelsystem.py
attesn: Updated this and I'll re-request review.

Scripts/datatypes/purpose.py
```diff
@@ -214,43 +214,34 @@ def __init__(self, specification, zone_data, resultdata):
...
-        self.resultdata.print_matrices(
-            self.aggregates, "aggregated_demand", self.name)
-        for mode in self.aggregates:
-            self.resultdata.print_data(
-                self.own_zone_demand[mode], "own_zone_demand.txt")
+    
+    @property
+    def generated_tours_all(self):
+        return pandas.Series(
+            sum(self.generated_tours.values()), self.zone_numbers, name=self.name)
```
zptro, line 227: I like the idea of feeding these `Series` object as a dict to `ResultData.print_data()` as you have done in `ModelSystem`. This means that the `Series.name` variable does not have any function anymore though. Should we remove it or could it be useful for something else in the future?

zptro (approved):

zptro (dismissed):

zptro (approved):

zptro (approved):

PR #17, opened 2024-06-07
Implement traversal analysis

zptro:
This should work for both sparser municipality-center network and for the full network. If run on the municipality-center network, some municipalities will have both the centroid and the freight terminal in the same zone number. In this case the demand on the matrix diagonal should be dis-aggregated according to some special rule.


Konnde (approved):
Tested to work as intended. Presumably overwrites latest output txt-file with intention. Tried with modification where commodity purpose key is included as additional function param for text output.

PR #19, opened 2024-06-13
Update long distance estimation with LOS data from 2016

attesn:
Primary survey data source is collected in 2016 and therefore LOS data should represent same base year. Until now estimation LOS price data has been collected in 2023. Here we update estimation with input data representing year 2016:
* Car cost and transit cost are transferred from 2023 to 2016 using consumer price indexes. 
* Transit time is updated with SYKE network data from year 2018.

### DIFF
--- a/Scripts/parameters/demand/hb_business_long.json
+++ b/Scripts/parameters/demand/hb_business_long.json
@@ -132,47 +132,47 @@
             "attraction": {},
             "impedance": {},
             "log": {
-                "logsum": 0.7201
+                "logsum": 0.72034
             },
             "individual_dummy": {}
         },
         "train": {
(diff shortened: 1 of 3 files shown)


attesn:
@EliasMarttunen Can you run validation for this model version, so we can see if mode shares have substantially changed to better or worse. 

I'd suggest that we do calibration in separate branch after merging this, so it can be reverted if needed (and some tests are performed to model without calibration).

attesn:
There seems to be errors in matrix processing here, so estimation results are not valid at this point.

attesn:
This Pull Request is on hold. We'll need to wait for new 2016 airplane cost data.

PR #20, opened 2024-06-19
Volume factors updated

EliasMarttunen:


Scripts/parameters/assignment.py
```diff
@@ -304,99 +304,69 @@
...
+        "pt": 1. / 0.1607,
+        "iht": 1. / 0.4886,
     },
     "transit_work": {
-        "aht": 1. / 0.445,
-        "pt": 1. / 0.103,
-        "iht": 1. / 0.414,
+        "aht": 1. / 0.709,
+        "pt": 1. / 0.3201,
+        "iht": 1. / 0.4886,
```
zptro, line 329: These seem suspiciously high.

Scripts/parameters/assignment.py
```diff
@@ -304,99 +304,69 @@
...
-    },
-    "long_d_bus": {
-        "aht": 1. / 0.478,
-        "pt": 1. / 0.109,
-        "iht": 1. / 0.405,
-    },
-    "airplane": {
-        "aht": 1. / 0.478,
-        "pt": 1. / 0.109,
-        "iht": 1. / 0.405,
```
zptro, line 359: I think it would be good to keep these after all, because they will be assigned on the hour-level in the regional sub-models and the model user may be interested in summing all link flows on day level.

Scripts/parameters/assignment.py
```diff
@@ -358,46 +393,6 @@
...
-    "van": {
-        "aht": 1 / 0.3,
-        "pt": 1 / 0.1,
-        "iht": 1 / 0.3,
-    },
-    "bus": {
-        "aht": 1 / 0.497, 
-        "pt": 1 / 0.090, 
-        "iht": 1 / 0.497,
-    },
```
zptro, line 400: We also still need "bus"

attesn (changes requested):
We'll need to update volume factors, as I didn't calculate these correctly from survey data.

PR #21, opened 2024-06-20
Embed parking cost in car cost matrix

attesn:
For future work with cost damping, cost sharing and values-of-time it would beneficial to have only single cost parameter in utility function. This pull request joins parking cost within car cost matrix. Parking costs are included at the destination of the tour, as it is assumed individuals do not have to pay to park at home

Activity time is weighted mean from National Travel Survey.  Share paying for parking is deducted by initially estimating model with separate parking cost parameter and then calculating ratio of `parking cost/ car cost`. 

Testing of cost and time damping are built on this branch.

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -74,11 +74,6 @@ class Purpose:
     def dest_interval(self):
         return slice(0, self.zone_data.nr_zones)
     
-    def add_destination_cost(self, mtx, vector):
-        """Add zonedata to matrix. """
-        mtx = numpy.add(mtx, vector)
-        return mtx
-
     def transform_impedance(self, impedance):
         """Perform transformation from time period dependent matrices
         to aggregate impedance matrices for specific travel purpose.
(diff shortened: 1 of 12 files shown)


Scripts/datatypes/purpose.py
```diff
@@ -73,6 +73,11 @@ def zone_numbers(self):
     @property
     def dest_interval(self):
         return slice(0, self.zone_data.nr_zones)
+    
+    def add_destination_cost(self, mtx, vector):
+        """Add zonedata to matrix. """
+        mtx = numpy.add(mtx, vector)
+        return mtx
```
zptro, line 80: I do not see the point in having this wrapper function, when you could simply write ` day_imp[mode][mtx_type] += park_cost` in `transform_impedance`.

zptro (changes requested):
Yes, this is a neat solution. Just one small remark.

Scripts/datatypes/purpose.py
attesn: This is nice solution!

zptro (approved):

PR #23, opened 2024-06-25
Cost damping in demand models

attesn:
Use linear and log cost formulation in daily tours model. Grocery model uses only log cost formulation.


attesn:
Estimation show marginally better fit for new model. Both linear and logarithmic time coefficient are negative and significant. There are small gains in log likelihood values. Derived values-of-time and elasticities are reasonable. 

Validation tests with trip length distributions and mode shares show that cost damping has only marginal impact on model issues predicting long trips. 

However, previous experiences seem to suggest that using linear and log cost should be the default specification and therefore I'd suggest that we merge.

zptro (approved):

PR #28, opened 2024-06-28
Change tour vectors into float32

zptro:
When 32 bit probability matrix is multiplied by 64 bit tours vector, the result has become a 64 bit matrix. This matrix is handled only temporarily, so the effect on memory consumption is small, but for large networks this might be critical.


samakinen (approved):
Looks good!

PR #29, opened 2024-07-02
Model data explorer and example analyses

attesn:
Add model data explorer to enable quick model data analyses. Main focus is to help model usage in case studies, but this could be beneficial for development as well - at least for matrix validation purposes.

Solution adds `data_explorer` module to model-system. This module can be accessed from notebooks/ui/etc. Explorer has two main components `ScenarioData` and `ScenarioComparison`. 

Model-system integration is not very well developed at the moment.


Scripts/data_explorer/scenario_data.py
```diff
@@ -0,0 +1,129 @@
...
+
+def read_mtx(file_path: Path, time_period: str, mtx_type: str, ass_class: str):
+    """
+    Read cost matrix from omx files.
+
+    Args:
+        time_period (str) : Time period
+        mtx_type (str) : Matrix type
+        ass_class (str) : Assignment class of model
+        zone_id (int) : Zone id for origin zone
```
zptro, line 30: This attribute is not in use

Scripts/data_explorer/scenario_comparison.py
```diff
@@ -0,0 +1,38 @@
...
+        self.zones = scenario0.zones
+        self.zone_ids = scenario0.zone_ids
+
+
+    def accessibility(self, tour_type, mode = None):
+        """
+        Get accessibility values of tour type (and mode).
+
+        Return : Series for tour type or mode.
+        """
```
zptro, line 27: This method does not do anything

Scripts/data_explorer/scenario_comparison.py
```diff
@@ -0,0 +1,38 @@
...
+        Args:
+            scenario0 : Scenario results of do-nothing scenario.
+            scenario1 : Scenario results of project scenario.
+        """
+        self.scenario0 = scenario0
+        self.scenario1 = scenario1
+        self.zones = scenario0.zones
+        self.zone_ids = scenario0.zone_ids
+
+
```
zptro, line 21: One empty row within classes.

zptro (changes requested):
I think the data_explorer module is well structured and useful. Just some small remarks.

Maybe the scripts in the Results folder could be moved to a separate repo in Bitbucket? If we want to share the user interface to these algorithms publicly, I think we should try to develop something more user friendly first. At the moment this should work well for internal use.

Scripts/data_explorer/scenario_comparison.py
attesn: Yep, I left some empty future features here on purpose. I thought that I'd get review for overall structure before finalizing them.

attesn:
> Maybe the scripts in the Results folder could be moved to a separate repo in Bitbucket? If we want to share the user interface to these algorithms publicly, I think we should try to develop something more user friendly first. At the moment this should work well for internal use.

Agreed! I'll develop notebooks here a bit further and then separate them to new repository. I'd think final solution could be a repository where we store the notebooks or scripts for each project (as there always are some custom features in visualizations etc).

attesn:
Solution here could be enhanced to more user friendly (without need to specify file names) and better integrated (using model-system data readers) if #30 is merged. Otherwise this should be ready.

PR #30, opened 2024-07-03
Refactor ZoneData and FileReader

attesn:
Purpose here is to simplify methods and make file reading easier in other methods (#29). ~~Base format of input data is same text files, but we'll do following changes:~~ @zptro  edit 2024/08/21: All input zone data moved to one file `zonedata.gpkg`. The mappings are in own columns named after the submodels.

* Full file names are used instead of file ends, so it is in line with results data files.
* Column names are changed to more self-explanatory versions `data_id `= `input_zone_id` and `zone_id`=`analysis_zone_id`.
* This `input_zone_id` index name is required when reading files.
* Separate methods to read zonedata and other data. I'd think this would make code more readable.
* Move aggregated input data to results, so all results are in similar zones.


Scripts/datahandling/zonedata.py
```diff
@@ -36,35 +36,32 @@ def __init__(self, data_dir: Path, zone_numbers: Sequence,
...
+        files = FileReader(data_dir, zone_mapping)
+        popdata = files.read_zonedata("population.tsv")
+        workdata = files.read_zonedata("workplaces.tsv")
+        incdata = files.read_zonedata("income_classes.tsv")
+        schooldata = files.read_zonedata("education.tsv")
+        parkdata = files.read_zonedata("parking_data.tsv")
+        sport_facilities = files.read_zonedata("sport_facilities.tsv")
+        buildings = files.read_zonedata("buildings.tsv")
+        self.transit_zone = files.read_other_data("transit_cost.tsv")
+        self.externalgrowth = files.read_zonedata("external_traffic.tsv")
```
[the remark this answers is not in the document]
attesn: I don't really see the benefits of using more complicated startswith syntax.  Anyways file versions need to be stored in separate folders and user can add metadata to beginning of file.

zptro (dismissed):

attesn:
Do we have a schedule for merging this? I've been waiting for some weeks already to progress with other tasks.

Scripts/datahandling/zone_variables.json
```diff
@@ -0,0 +1,35 @@
...
+    "share_age_50-64": "weighted_avg_population",
+    "share_age_65-99": "weighted_avg_population",
+    "sh_income_0_19": "weighted_avg_population",
+    "sh_income_20_39": "weighted_avg_population",
+    "sh_income_40_59": "weighted_avg_population",
+    "sh_income_60_79": "weighted_avg_population",
+    "sh_income_80_99": "weighted_avg_population",
+    "sh_income_100_": "weighted_avg_population",
+    "workplaces": "sum",
+    "sports_in": "sum",
```
attesn, line 15: "sport_facility_indoor"
"sport_facility_outdoor"

Scripts/datahandling/zone_variables.json
```diff
@@ -0,0 +1,35 @@
...
+    "sports_out": "sum",
+    "ski_resort": "max",
+    "area_education": "sum",
```
attesn, line 18: "area_education_building"
"area_leisure_building"

Scripts/datahandling/zone_variables.json
```diff
@@ -0,0 +1,35 @@
...
+    "area_leisure": "sum",
+    "wrk_sh_service": "weighted_avg_workplaces",
+    "wrk_sh_shop": "weighted_avg_workplaces",
+    "wrk_sh_hospitality": "weighted_avg_workplaces",
+    "wrk_sh_recreation": "weighted_avg_workplaces",
+    "avg_park_cost": "mean",
+    "avg_park_time": "mean",
+    "comprehensive": "sum",
```
attesn, line 26: "students_comprehensive"
"students_upper_secondary"
"students_higher_education"

Scripts/datahandling/zone_variables.json
```diff
@@ -0,0 +1,35 @@
...
+    "upper_secondary": "sum",
+    "higher_education": "sum",
+    "car_density": "mean",
+    "municipality_agg": "first",
```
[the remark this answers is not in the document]
zptro: We need some way of identifying these aggregation variables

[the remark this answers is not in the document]
samakinen: My vote still goes to separating the aggregation type and share definitions on different levels. Something like in our internal model-data-integration branch:
```
  "aggregation": "AREA_SHARE",
  "columns":
  [
      {
          "result": "population",
          "column": "pop_total",
          "shares":
          {
              "sh_7-17": "pop_sh_7-17",
              "sh_1829": "pop_sh_1829",
              "sh_3049": "pop_sh_3049",
              "sh_5064": "pop_sh_5064",
              "sh_65-": "pop_sh_65-"
          }
      },..
```
That helps to keep the data and the logic separate and makes the configuration files more explicit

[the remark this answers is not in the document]
zptro: @samakinen I approve this and will make the appropriate changes. That is however another question than the special result aggregation attributes.

zptro:
@samakinen Could you review this again now?

samakinen (dismissed):
I think we could merge this now and extend/refactor the aggregations later to work with spatial data as well.

zptro:
I tested this with the full-scale data and had to make fixes related to the dropped aggregation level "area". I now replaced it with "county" as a major aggregation level.

I also made `log` module `logger` specific for that module, so that it does not have to struggle with logs from other packages, such as `fiona`.

Scripts/utils/read_csv_file.py
```diff
@@ -1,154 +1,42 @@
...
+    pandas.DataFrame
+    """
+    if not path.exists():
+        msg = f"Path {path} not found."
+        raise NameError(msg)
+    header = None if squeeze else "infer"
+    data: pandas.DataFrame = pandas.read_csv(
+        path, delim_whitespace=True, keep_default_na=False, 
+        na_values="", comment='#', header=header)
+    if "node_label" in data.columns:
```
samakinen, line 29: I'm not sure why this is here but I think "node_labels" should not be a hard-coded string in the data loader function. Maybe this could be an optional "index_column" argument that is then defined in the relevant configuration file or calling function.

zptro: Yes, I agree. I put this back into the code to make it work as intended, but overall I have done no changes to this method. My plan is to completely reorganize these "other" data in an upcoming PR and put them all in one json-format file. After that we have the possibility to refer directly to one gpkg file (for zone data) and one json file (for data concerning travel costs) in the UI.

samakinen (approved):

zptro:
Will I ever be able to merge this? Now I just merged main though.

samakinen (approved):

PR #32, opened 2024-07-07
Long distance demand model updates

attesn:
Collection of small changes to enhance long distance modelling:
* Use updated 2023 impedance with Lapland flight connections.
* Add ski resorts to attraction variables.
* Use car density in car driver and passenger utility functions.


attesn:
Update yields likelihood gains on estimation and better tour length distributions. However, strangely car density update leads to worse mode share results on model-system run.

Scripts/parameters/demand/hb_business_long.json
```diff
@@ -132,47 +132,51 @@
...
             "individual_dummy": {}
         },
         "car_work": {
-            "constant": 1.32378,
+            "constant": -0.14918,
             "generation": {},
-            "attraction": {},
+            "attraction": {
+                "car_density": 3.62549
+            },
```
zptro, line 164: This should be on in the generation group for clarity. But it should not affect results in this kind of model structure.

zptro (approved):

attesn:
@EliasMarttunen Could you document validation runs between this and main branch? New word document with model parameters (Työtila / Henkilöliikenne / Estimointi / Alogit_summary.xlsx), mode share and tour length distributions.

zptro (approved):

attesn:
Apparently there was errors in LOS matrices, so we'll need to update estimation before merging.

PR #33, opened 2024-08-06
Do demand matrix aggregations for all available mappings

zptro:
Instead of one hard-coded aggregation level, all aggregation levels found in the .agg file will be used for printing separate files.

Also fix matrix direction, which has been wrong for a while.


attesn (approved):
This is really good addition for model validation runs!

PR #34, opened 2024-08-06
Enforce long-distance mode for long-distance assignment classes

zptro:
This may increase model run times somewhat, but removes need for hard-coded airplane cost threshold.

- Also remove count_zone_boardings artefacts inherited from helmet
- Update docstrings

This is the same PR that was in Bitbucket earlier.


juliuskroetzl (approved):

juliuskroetzl (approved):

attesn (approved):

attesn:
Can be merged together with #259

attesn:
@juliuskroetzl Can you re-review changes? Dev branch is now estimated with matrices produced by this branch, so it would be good to merge.

PR #37, opened 2024-08-08
Use branch of industry in home-based work demand models

attesn:
Branch of industry could affect mode choice in addition to job accessibility. Hypothesis here is that office workers have flexible nine to five working hours, which makes it easier to use public transport when compared to "blue collar" workers. 

Survey data does not include occupational classification, so we are using share of office jobs as attraction variable in zonal level. Classification of "office" jobs:

0 Sotilaat = Other
1 Johtajat = Office
2 Erityisasiantuntijat = Office
3 Asiantuntijat = Office
4 Toimisto- ja asiakaspalvelutyöntekijät = Office
5 Palvelu- ja myyntityöntekijät = Other
6 Maanviljelijät, metsätyöntekijät ym. = Other
7 Rakennus-, korjaus- ja valmistustyöntekijät = Other
8 Prosessi- ja kuljetustyöntekijät = Other
9 Muut työntekijät = Other


attesn:
image

National level does not show any significant change. but on subarea level there is modest improvement. The change is most significant in Helsinki downtown, where share of transit trips increases when branch of industry is part of model. However, this is clearly not enough to explain the underestimation of transit mode shares in large city centers.

zptro (approved):

PR #40, opened 2024-08-13
Split purpose and mode names in `print_matrices`

### LINKED ISSUE #38: Split purpose and mode to separate columns in `print_matrices`
https://github.com/Traficom/lem-model-system/blob/231ae55a78271e5340848621725f33d5c3706c18/Scripts/datahandling/resultdata.py#L118C64-L118C68

Aggregated demand result files have combined index column of purpose + mode, which causes extra work in analysis. Would it be possible to keep these in separate columns? A minimal solution would be to use some other character than undescore `"_"` to separate purpose and mode, as purpose is usually multipart name with underscores.

attesn:
Closes #38

### DIFF
--- a/Scripts/datahandling/resultdata.py
+++ b/Scripts/datahandling/resultdata.py
@@ -115,9 +115,7 @@ class ResultsData:
                     sheet_name=f"{description}_{key}")
         # Create text file
         stacked_matrices = pandas.concat(
-            {key: df.stack() for key, df in data.items()},
-            names=["mode", "orig", "dest"])
+             {(description, key): df.stack() for key, df in data.items()},
+            names=["purpose", "mode", "orig", "dest"])
         stacked_matrices.name = "nr_tours"
-        stacked_matrices = stacked_matrices.reset_index()
-        stacked_matrices.insert(0, 'purpose', description)
-        self.print_concat(stacked_matrices.set_index("purpose"), filename + ".txt")
+        self.print_concat(stacked_matrices, filename + ".txt")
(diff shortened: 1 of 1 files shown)


attesn:
Didn't include it here, but I'd remove the Excel files for simplicity.

Scripts/datahandling/resultdata.py
```diff
@@ -115,7 +115,9 @@ def print_matrices(self,
                     sheet_name=f"{description}_{key}")
         # Create text file
         stacked_matrices = pandas.concat(
-            {f"{description}_{key}": df.stack() for key, df in data.items()},
-            names=["purpose", "orig", "dest"])
+            {key: df.stack() for key, df in data.items()},
+            names=["mode", "orig", "dest"])
```
zptro, line 119: I played a bit with `pandas.concat` and realized that you can achieve the same result like this:

```suggestion
            {(description, key): df.stack() for key, df in data.items()},
            names=["purpose", "mode", "orig", "dest"])
```

Scripts/datahandling/resultdata.py
```diff
@@ -115,7 +115,9 @@ def print_matrices(self,
...
         stacked_matrices.name = "nr_tours"
-        self.print_concat(stacked_matrices, filename + ".txt")
+        stacked_matrices = stacked_matrices.reset_index()
+        stacked_matrices.insert(0, 'purpose', description)
+        self.print_concat(stacked_matrices.set_index("purpose"), filename + ".txt")
```
zptro, line 123: These additions are then unnecessary

```suggestion
        self.print_concat(stacked_matrices, filename + ".txt")
```

zptro (changes requested):

zptro (approved):

PR #41, opened 2024-08-15
Update long distance estimation with 2016 impedance

attesn:
New 2016 impedance matrices used for estimation. Previous attempts had errors in aiplane transit time function, which caused bias in impedance data and parameters. Also Lapland flight connections with zero travel time was updated in matrices.


attesn:
@EliasMarttunen You can approve if validation mode shares seem reasonable.

EliasMarttunen (approved):

PR #51, opened 2024-08-22
Add printing of mode boardings per municipality

zptro:


attesn (approved):

attesn:
Looks good!

attesn:
One idea could be to print "extra_nodes_validation" table as well, if those attributes are found from network. Then network validation would be very easy (and maybe more often done in model use-cases).

attesn (approved):

zptro:
Messed up my branches, sorry!

PR #53, opened 2024-08-26
Short tours calibration

EliasMarttunen:
Short tours calibration parameters after one calibration round.

### DIFF
--- a/Scripts/parameters/demand/hb_edu_basic.json
+++ b/Scripts/parameters/demand/hb_edu_basic.json
@@ -185,6 +185,20 @@
                 "logsum": 0.57426
             },
             "individual_dummy": {}
+        },
+        "calibration": {
+            "car_pax": {
+                "constant": 0.023335131
+            },
+            "transit_work": {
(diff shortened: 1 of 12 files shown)


zptro (changes requested):
This PR fails two tests. The first one is obvious; we have a test that makes sure the developer does not unintentionally change model results:

```
# Check that model result does not change
        self.assertAlmostEquals(
            model.mode_share[0]["car_work"] + model.mode_share[0]["car_leisure"],
>           0.19920193721251422)
E       AssertionError: 0.25385090706936597 != 0.19920193721251422 within 7 places (0.05464896985685175 difference)
```

So in this case when you *want to* change model results, you need to change  0.19920193721251422 into 0.25385090706936597 in `Scripts/tests/integration/test_models`.

The other error is a bug in one of the other tests that I can try to fix.

attesn (approved):

attesn:
@EliasMarttunen You can merge changes. Probably good to choose "Squash and merge" option.

PR #54, opened 2024-08-27
Introduce json parameter file as alternative to command-line parameters

zptro:
- Generalize `config` module to handle other JSON-format config files than `dev-config.json` and add reverse (i.e., JSON dump function)
- Unify JSON and command-line parameter names, so that no explicit matching is needed
- Add new `--json` parameter in `lem.py`
- Change parameter dump in `log` to JSON format that it can be used directly as input to other model run


Scripts/create_emme_project.py
```diff
@@ -118,47 +118,39 @@ def create_emme_project(args):
...
         action="store_true",
-        default=config.SEPARATE_EMME_SCENARIOS,
         help="Using this flag enables saving network time-period specific results in separate EMME scenarios."),
     parser.add_argument(
         "--first-scenario-id",
         type=int,
-        default=config.FIRST_SCENARIO_ID,
         help="First EMME project scenario ID"),
+    parser.set_defaults(
+        **{key.lower(): val for key, val in config.items()})
```
zptro, line 151: This could be simplified if we change the parameter names to lower keys in `config`.

Scripts/create_emme_project.py
```diff
@@ -118,47 +118,39 @@ def create_emme_project(args):
...
     args = parser.parse_args()
-
-    args_dict = vars(args)
-    for key in args_dict:
-        log.debug("{}={}".format(key, args_dict[key]))
+    log.initialize(args)
+    log.debug(utils.config.dump(vars(args)))
```
zptro, line 154: We could of course print it to a separate file as well.

Scripts/lem.py
```diff
@@ -171,28 +171,29 @@ def main(args):
...
+    parser.add_argument(
+        "--json",
+        type=str,
+        help="Read parameters from file, override command-line and dev-config.json arguments",
+    )
     # Logging
     parser.add_argument(
         "--log-level",
         choices={"DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"},
-        default=config.LOG_LEVEL,
```
samakinen, line 178: We could save all possible parameters in a list of NamedTuples. Then we could just iterate that list here and set the parser arguments:
`for a in config.arguments: parser.add_argument(a.cmd_flag, type=a.type, default=a.default, help=a.help)`
and use the same list in other places like the config.py

Scripts/utils/config.py
```diff
@@ -39,37 +60,42 @@ class Config:
         value : str/bool/int/float
             Parameter value
     """
+    c = Config()
+    c.update({
+        "LEM_VERSION": None,
```
samakinen, line 65: LEM_VERSION is still referred as HELMET_VERSION in some parts of the code

Scripts/create_emme_project.py
samakinen: I think it would be a good idea to write the whole configuration in the result directory as a json file. That would make it very easy to repeat the calculation afterwards if needed.

Scripts/create_emme_project.py
samakinen: Maybe it's better to leave this as it is. It's not that much extra code and can avoid some potentially hard to find errors if someone mistypes the case of the key.

samakinen (approved):
Looks good and makes it much easier to debug different configurations or repeat previously run calculations.

samakinen (approved):
Looks good, I think we can merge it.

PR #55, opened 2024-08-28
Add cost sharing of car driver and passenger to demand models

attesn:
New estimation of short and long tours with cost sharing between car driver and passenger. Utility function formulation is based on Fox et al. 2009 and Rich and Fox 2024.

Model-system implementation is now simple modification to impedance transformer method. Same time all impedance tranformations were collected to new cost parameters.

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -51,8 +51,8 @@ class Purpose:
         self.area = specification["area"]
         self.impedance_share = specification["impedance_share"]
         self.demand_share = specification["demand_share"]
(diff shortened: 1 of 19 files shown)


attesn:
Full model test run exits process after starting demand calculation. There was no error messages. Memory issues?

zptro:
I cannot figure out why this would cause memory issues. But the problem you describe indeed suggest we have memory issues.

attesn:
@zptro Me neither, so I tried again and model worked without any problems. Not sure if I got some background process or something else causing the error on previous time. So I'll revert previous comment.

zptro (approved):

PR #56, opened 2024-08-29
Add log messages to monitor demand model progress

attesn:
Demand model can run > hour without any progress reports, so add few purpose specific messages.


zptro (approved):
Looks good to me!

PR #57, opened 2024-09-06
Release memory more efficiently in logit models

zptro:
A few things to lower maximum memory consumption by 2-3 GB in short-trip "koko_suomi" mode.

- Run accessibility calculations before creation of probability matrices, so that impedance matrices can be deleted before probability matrix creation
- Make temporary matrices strictly local, so that memory is cleared immediately when usage stops
- Use `dict.pop()` more

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union, cast
+from copy import copy
 from collections import defaultdict
 import numpy # type: ignore
 import pandas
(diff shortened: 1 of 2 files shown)


Scripts/datatypes/purpose.py
```diff
@@ -278,10 +279,10 @@ def calc_prob(self, impedance, is_last_iteration):
                 Type (time/cost/dist) : numpy 2d matrix
         """
         purpose_impedance = self.transform_impedance(impedance)
-        self.prob = self.model.calc_prob(purpose_impedance)
         if is_last_iteration and self.name[0] != 's':
             self.accessibility_model.calc_accessibility(
-                purpose_impedance)
+                copy(purpose_impedance))
+        self.prob = self.model.calc_prob(purpose_impedance)
```
zptro, line 285: By copying the dict, we can pop the matrices in them on usage, hence depleting both dicts. When the reference count for the matrices drops to zero (in `calc_prob`), memory is released.

attesn (approved):

PR #58, opened 2024-09-09
Update demand model with new car cost

attesn:
Use only fuel costs as car kilometer based cost. Maintenance cost does not seem to improve model fit and fuel cost is more in line with models that we use to validate results.


zptro (commented):
Does this change improve/preserve VoT?

attesn:
> Does this change improve/preserve VoT?

Seems that for most models VoT is higher with new specification. Below is example of home-based work tours.

image

PR #59, opened 2024-09-10
First draft of relaxed time periods

zptro:
This PR enables producing evening transit LOS and using them in demand model.

As a separate step we should try model estimation without the "aht" LOS, to see if that time period can be dropped in demand calculation loop.

~~As suggested in this PR, we should also test replacing bike and walk distance with time and to entirely drop distance matrix from demand calculation. Mostly time is a proxy for distance, but when running the model for archipelago areas, there will be a difference in how people use ferries in walk mode.~~ To be able to run the model with more LOS matrices (evening added) it was necessary to remove distance at least for the modes that do not use it in demand calculation.

TODO: Matrices and extra attributes in EMME are currently a mess! In a later PR, I will try to refactor all of that.


attesn:
> As suggested in this PR, we should also test replacing bike and walk distance with time and to entirely drop distance matrix from demand calculation. Mostly time is a proxy for distance, but when running the model for archipelago areas, there will be a difference in how people use ferries in walk mode.

Here I disagree. Archipelago areas and ferries are such small portion of total traffic that I would prioritize other development work. On quick look there is < 0,2 % of population living in these areas, so we should get the 80 percent correct before starting to fix these.

attesn (commented):
Looks good! However, I don't like the idea of including distance matrix drop to this update. Archipelago areas and ferries are such small portion of total traffic that other development work should be prioritized. On quick look there is < 0,2 % of population living in these areas and we many many open issues with the 80 percent.

Maybe we should also test the run with off peak period before merging? Therefore I didn't approve changes yet.

zptro:
> Looks good! However, I don't like the idea of including distance matrix drop to this update. Archipelago areas and ferries are such small portion of total traffic that other development work should be prioritized. On quick look there is < 0,2 % of population living in these areas and we many many open issues with the 80 percent.
> 
> Maybe we should also test the run with off peak period before merging? Therefore I didn't approve changes yet.

I restored distance matrices for walk and bike. To be able to run the model with more LOS matrices (evening added) it was necessary to remove distance at least for the modes that do not use it in demand calculation.

Scripts/tests/test_data/Base_input_data/Matrices/uusimaa/demand_ilt.omx
attesn, line 1: I think this file has wrong shape (12, 12)?

attesn:
Työtila matrices have suffix "it" and here it seems to be "ilt". Which is preferred?

zptro:
> Työtila matrices have suffix "it" and here it seems to be "ilt". Which is preferred?

I will change the name to "it" everywhere, but aside from that, this PR seems to be finished.

attesn:
@zptro I think we'll need transit assignment for `pt` with free flow based speeds? Estimation was done only with `it` matrix updated.

zptro:
> @zptro I think we'll need transit impedance matrices for `pt` with free flow based speeds (to match estimation with this PR model-system)? Estimation inluded was done only with updated `it` matrix.

Done.

attesn (dismissed):
@zptro Ready! However it seems that my review is not valid as I added new commits.

zptro:
> @zptro Ready! However it seems that my review is not valid as I added new commits.

So I by-passed the rules.

PR #61, opened 2024-09-13
Update volume factors for long-dist transit

zptro:
This is a technical solution for restoring whole-day passenger volumes for long-distance transit in short trip (submodel) assignment. These must be updated if demand shares for external traffic is updated here: https://github.com/Traficom/lem-model-system/blob/main/Scripts/parameters/departure_time.py#L27. Error in passenger volumes, which this fixes, is -18%.


zptro:
This error has been cancelled out by another error in tour generation.

attesn (changes requested):
Handled in #70

zptro:
@attesn No, this is in fact more relevant now when tour generation (which has until now been overestimated) has been fixed.

attesn:
> @attesn No, this is in fact more relevant now when tour generation (which has until now been overestimated) has been fixed.

But open tours were already taken into account when calculating the new generation parameters.

zptro:
> > @attesn No, this is in fact more relevant now when tour generation (which has until now been overestimated) has been fixed.
> 
> But open tours were already taken into account when calculating the new generation parameters. Or did I misunderstood the purpose here?

This is not about open/closed tours. This is about the fact that volume factors for long-distance modes has been temporally taken from Helmet mode "transit", while demand shares have been taken from Helmet external transit. These do not add up correctly. 

This update is "technical" in the sense that we do not need to know the true volume factors to restore whole-day volumes when we know what demand shares have been used to split them in the first place. In CBA, however, the volume factors also affect weighting of LOS, so there it is more important to have correct volume factors.

PR #62, opened 2024-09-13
Small adjustment of logging

zptro:


Scripts/datatypes/purpose.py
```diff
@@ -296,6 +295,7 @@ def calc_prob(self, impedance, is_last_iteration):
             self.accessibility_model.calc_accessibility(
                 copy(purpose_impedance))
         self.prob = self.model.calc_prob(purpose_impedance)
+        log.info(f"Mode and dest probabilities calculated for {self.name}")
```
zptro, line 298: This is part of the first heavy loop of the demand model and `calc_prob` is the most time-consuming step in that loop, so it makes sense to log when that step is done.

Scripts/modelsystem.py
```diff
@@ -165,7 +165,6 @@ def _add_internal_demand(self, previous_iter_impedance, is_last_iteration):
                 if purpose.dest != "source":
                     for mode in demand:
                         self.dtm.add_demand(demand[mode])
-                    log.info(f"Time period matrices calculated for {purpose.name}")
```
zptro, line 168: This is part of the second heavy loop of the demand model. But the most time-consuming step of that loop, `calc_demand` is already logged, so this does not add much useful information.

PR #65, opened 2024-09-17
Long tours parameters with mode enforcement and fixed value-of-time

attesn:
This PR includes changes from #34 and estimation with fixed value-of-times for long tours.


attesn:
@EliasMarttunen Elasticity tests could be run for this PR. You need to use impedance matrices from "Työtila / 08 / Vastusmatriisit / koko_suomi_free_flow_mode_enforcement".

PR #66, opened 2024-09-17
Update long distance parameters with fixed VoT and mode enforcement

attesn:


attesn:
@EliasMarttunen Elasticity tests could be run for this PR. You need to use impedance matrices from "Työtila / 08 / Vastusmatriisit / koko_suomi_free_flow_mode_enforcement".

Scripts/datatypes/purpose.py
```diff
@@ -164,6 +164,15 @@ def transform_impedance(self, impedance):
...
                         pass
+        for name in cost.gen_cost:
+            if name == self.name:
+                for mode in cost.gen_cost[name]:
+                    try:
+                        vot = cost.gen_cost[name][mode]
+                        day_imp[mode]["gen_cost"] = (day_imp[mode]["time"] / 60 * vot 
+                                                     + day_imp[mode]["cost"])
+                    except KeyError:
+                        pass
```
zptro, line 175: I think we should implement time and cost parameters as regular json model parameters instead. For the model user it is not relevant whether the model is estimated with fixed value of time or not. Hence, this extra script does not add any value in my opinion.

attesn: Could be, but I think we can sort this out later if we want to merge. This is more convenient way when testing demand model specification.

attesn:
Based on the discussion in demo meeting and results from validation, I think it is not feasible to merge this PR. Using imported values should lead to better model (mode shares, elasticities), which was not the case here.

PR #69, opened 2024-09-20
Refactor purpose calling slightly

zptro:
Taking this further would require rethinking of how `DepartureTimeModel` is handled, which may or may not be a good idea at some point. But this refactoring was a low-hanging fruit that @samakinen pointed out.


zptro:
@samakinen Is this ok now?

samakinen:
> @samakinen Is this ok now?

Yes, looks much better now.

samakinen (approved):

PR #70, opened 2024-09-20
Use submodel specific long tour rates

attesn:
Using submodel specific rates gives better results than demographics (age, income). Both cannot be used in tour rates, because observations are limited. This should be considered as temporary solution, using probability models would allow using both demographics and accessibility.

image


Scripts/datahandling/zonedata.py
```diff
@@ -100,6 +100,8 @@ def __init__(self, data_dir: Path, zone_numbers: Sequence,
         self["outside_municipality"] = ~within_municipality
         for dummy in ("Helsingin_kantakaupunki", "Tampereen_kantakaupunki"):
             self[dummy] = self.dummy("subarea", dummy)
+        for key in data["aggregate_results_submodel"].unique():
+            self["population_" + key] = (data["aggregate_results_submodel"] == key) * pop
```
zptro, line 104: This could also be written as:

```suggestion
            self["population_" + key] = self.dummy("submodel", key) * pop
```

zptro (approved):
Looks good!

zptro (approved):

PR #89, opened 2024-10-25
Merging updated test network data into main

Konnde:
Created new test network that has been tested to run through with latest main changes. Updated emme_only and without emme test_assignment python modules to cover changes in test network and main.


zptro (changes requested):
Thanks for this! For the tests to pass, we still need to update the matrices (tests/test_data/base_input_data and tests/test_data/results/test) and some small details in the tests themselves. Do you want to give it a try or should I do it?

Konnde:
I can take this in order to learn more about the system! After glancing the failing build, indeed it seems that I forgot to update the matrices to match network centroids. I'll make the necessary updates.

zptro (dismissed):
This seems to work really well, great job!

zptro (approved):

PR #90, opened 2024-10-28
Add import of line data to `MockProject`, so that tests work as intended

zptro:


Konnde:
after chatting, I added these updates into local repo in order to fix testing fares in unit tests / test_assignment.py which was not raising an error previously. This fix was able to solve the problem. We can now close this PR and the change will come through along my updates to new test network.

PR #91, opened 2024-10-30
New demand parameters with fixed nests

attesn:
Estimation is using dummy nests structure to avoid memory limits caused by high number of alternatives. This means that additional dummy level of hierarchy is added to nests and logsum coefficient is fixed to one (causing estimation to be equal to two level hierarchy). However, wrong logsum coefficient has been fixed in previous estimations.

This causes significant changes in estimation results, which are presented here. 

There is some work to do before merging:
1. HB Work, HB University and WB Other models use `DestMode` structure , which is not working. :white_check_mark:
2. New Accessibility model for `DestMode` model, so generation model for university purpose can be re-estimated. :white_check_mark:
3. HB University generation parameters need to be updated with new accessibility values. :white_check_mark:
4. Validation report with elasticities. :white_check_mark:


attesn:
Home-based Visit model shows some strange car driver mode shares. Also car passenger mode is systematically underestimated.
mode_share

zptro (approved):
So the modal split changes a lot in the integration test, which I assume is partially because calibration parameters were removed. So as a separate step, I guess this will be re-calibrated?

attesn:
> So the modal split changes a lot in the integration test, which I assume is partially because calibration parameters were removed. So as a separate step, I guess this will be re-calibrated?

Yep, we have few other updates coming up soon, so we'll wait for them before re-calibration. However, its better to have new features in separate Pull Requests.

PR #92, opened 2024-10-30
Fix `DestModeModel`

zptro:
- Fixed deprecated `_calc_dest_util` method call in `DestModeModel`
- Fixed logit test to work also with `DestModeModel`
- Added `DestModeModel` logsum that can be used by tour generation model
- Added placeholder for accessibility measures and printing

Please note that there are no mode-specific logsums in the `DestModeModel` (as mode choice is on the OD level), so they cannot be used as a zone variable without some kind of not-yet-existing magic.

### DIFF
--- a/Scripts/models/logit.py
+++ b/Scripts/models/logit.py
@@ -607,8 +607,16 @@ class DestModeModel(LogitModel):
                 Choice probabilities
         """
         mode_expsum = self._calc_mode_util(impedance)
-        logsum = {"logsum": mode_expsum}
-        dest_expsum, dest_exps = self._calc_dest_util("logsum", logsum)
+        dest_exps = self._calc_dest_util("logsum", {"logsum": mode_expsum})
+        try:
+            dest_expsum = dest_exps.sum(1)
+        except ValueError:
+            dest_expsum = dest_exps.sum()
+        logsum = pandas.Series(
+            numpy.log(dest_expsum), self.purpose.zone_numbers,
+            name=self.purpose.name)
+        self.accessibility = {"all": logsum}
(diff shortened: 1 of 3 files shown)


attesn (changes requested):
Testing with full model system ends with `ValueError`: `Matrix car_work in file demand_aht.omx contains NA values`. This probably is due to the generation model as it happes after HB Higher Edu calculations.

attesn:
> Testing with full model system ends with `ValueError`: `Matrix car_work in file demand_aht.omx contains NA values`. This probably is due to the generation model as it happes after HB Higher Edu calculations.

When tour combination parameters are removed the problem stays, which means that NA values are produced by destination-mode choice model.

Scripts/models/logit.py
```diff
@@ -607,15 +607,26 @@ def calc_prob(self, impedance):
...
+            dest_expsum = dest_exps.sum()
+        logsum = pandas.Series(
+            numpy.log(dest_expsum), self.purpose.zone_numbers,
+            name=self.purpose.name)
+        self.accessibility = {"all": logsum}
+        self.zone_data._values[self.purpose.name] = logsum
         prob = {}
         dest_prob = dest_exps.T / dest_expsum
         for mode in self.mode_choice_param:
             mode_prob = (self.mode_exps[mode] / mode_expsum).T
```
attesn, line 623: This produces NAs, I think `mode_prob = numpy.divide(self.mode_exps[mode], mode_expsum, out=numpy.zeros_like(self.mode_exps[mode]), where=mode_expsum!=0).T` fixes this.

attesn (approved):

PR #93, opened 2024-10-30
Add freight assignment test

zptro:


Konnde (approved):
Tested this and it goes through. We'll merge this as it's own test and include without emme freight demand testing as it's own pull request.

PR #95, opened 2024-11-01
Improve index in car_time table, so that weird qoute-mark handling can be removed from resultdata

zptro:
This removes the strange double spaces in `str` values.


attesn (approved):
Thanks for fast work on this! I dont't have working assignment model to do model run, but code seems ok.

PR #97, opened 2024-11-01
Feat/setup freight tests

Konnde:
Implemented without emme freight demand testing. Freight test variables are now also found in test datas with added test freight impedance matrices. Additionally merged emme only freight assignment testing from it's own branch into this so we can drop that branch.

How does this look?


Scripts/tests/test_data/Network/netfield_transit_lines_test.txt
zptro, line 1: Why did you change this file?

Scripts/utils/freight_costs.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def calc_road_cost(unit_cost_parameters: Dict[str, Dict],
+                   distance: numpy.ndarray,
+                   time: numpy.ndarray,
+                   toll_cost: numpy.ndarray = 0) -> numpy.ndarray:
+    mode_cost = {}
+    for mode, parameters in unit_cost_parameters.items():
+        time_cost = time * parameters["time_based_cost"]
+        dist_cost = distance * parameters["distance_based_cost"]
```
zptro, line 12: Using the same logic as elsewhere, you could consider having matrices in a dict with keys "time" and "dist" and parameters named exactly in the same way, so we do not need to have the keys hard-coded here.

Scripts/utils/freight_costs.py
```diff
@@ -0,0 +1,64 @@
...
+        time_cost =  parameters["time_based_cost"] * time
+        dist_cost = distance * parameters["distance_based_cost"]
+        rail_cost = ((time_cost+dist_cost+toll_cost)
+                     / parameters["average_load"]*2
+                     + parameters["wagon_yearly_cost"])
+        mode_cost[mode] = rail_cost + (parameters["terminal_cost"]*2)
+    rail_cost = mode_cost["diesel_train"]
+    rail_aux_cost = numpy.where(
+        (aux_distance > (distance*2)) | (distance == 0), numpy.inf,
+        calc_road_cost(road_unit_costs, aux_distance, aux_time))
```
zptro, line 39: By turning this into a function, you could reuse it for ships.

Scripts/tests/unit/test_freight_demand.py
```diff
@@ -0,0 +1,80 @@
...
+                    "trailer_truck": freight_cost["trailer_truck"],
+                },
+                "freight_train": {
+                    "electric_train": freight_cost["electric_train"],
+                    "diesel_train": freight_cost["diesel_train"]
+                },
+                "ship": {
+                    "other_dry_cargo": {"4m": freight_cost["other_dry_cargo"]["4m"]} 
+                }
+            }
```
zptro, line 57: Try to structure `freight_cost` so that you do not need to do this restructuring here.

zptro (changes requested):
Nice work!

As assignment tests are not changed in this PR, it is unclear what the meaning of the suggested changes in freight assignment is. Could you drop them from this PR to make it more concise? I think it is better to merge my suggested assignment-test changes separately, and from there continue the development of freight assignment as a separate PR, where you can explain and motivate those changes.

Scripts/datatypes/purpose.py
```diff
@@ -538,3 +538,37 @@ def calibrate(spec: dict, calib_spec: dict):
...
+        Parameters
+        ----------
+        impedance : dict
+            Mode (truck/train/...) : dict
+                Type (time/cost/dist) : numpy 2d matrix
+
+        Return
+        ------
+        datatypes.demand.Demand
+            Freight mode demand matrix for whole day
```
zptro, line 567: You return a dict and not this class

Scripts/tests/unit/test_freight_demand.py
```diff
@@ -0,0 +1,80 @@
...
+            with open(os.path.join(PARAMETERS_PATH, file_name), 'r') as file:
+                comm_params = json.load(file)
+                comm = file_name.split("_")[0]
+                purposes[comm] = FreightPurpose(comm_params, zonedata, resultdata)
+        with open(TEST_DATA_PATH / "costdata.json") as file:
+            costdata = json.load(file)
+        
+        time_impedance = omx.open_file(TEST_MATRICES / "freight_time.omx", "r")
+        dist_impedance = omx.open_file(TEST_MATRICES / "freight_dist.omx", "r")
+        comm_map = {"marita": "mameka", "kalevi": "mameka"}
```
zptro, line 41: This map would be good to put somewhere else.

Scripts/assignment/freight_assignment.py
```diff
@@ -48,8 +49,15 @@ def save_network_volumes(self, commodity_class: str):
             spec["on_links"]["aux_transit_volumes"] = '@a_' + attr_name
             self.emme_project.network_results(
                 spec, self.emme_scenario, ass_class)
+        attr_name = (commodity_class + "truck")[:17]
+        link_costs = {"truck": 1.0, "semi_trailer": 1.0, "trailer_truck": 1.0}
+        carspec = CarSpecification(self.extra, self.emme_matrices, link_costs)
```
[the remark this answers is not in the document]
Konnde: Now that we have merged freight assignment testing into main, I can drop these changes from this PR.

Scripts/datahandling/zone_variables.json
```diff
@@ -33,7 +33,11 @@
         "area_university_buildings",
         "students_comprehensive",
         "students_upper_secondary",
-        "students_higher_education"
+        "students_higher_education",
+		"gen_marita",
+		"gen_kalevi",
+		"att_marita",
+		"att_kalevi"
```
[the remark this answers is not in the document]
Konnde: This indentation error doesn't appear in local, interesting. Will find out what is happening here for next review.

Scripts/utils/freight_costs.py
Konnde: Will refactor cost calculations and for all freight modes and their unit cost parameter json files to improve readability.

Scripts/tests/unit/test_freight_demand.py
Konnde: I think adding upper mode level and sub mode level into ```freight_cost``` will improve this, e.g. commodity : freight_train : electric_train.

Scripts/datatypes/purpose.py
Konnde: Faulty docstring, will amend this section.

Scripts/tests/unit/test_freight_demand.py
Konnde: Suggesting an addition of a new ```commodity.py``` file into parameters directory where conversion between commodity classes are specified.

Scripts/datahandling/zone_variables.json
```diff
@@ -33,7 +33,11 @@
         "area_university_buildings",
         "students_comprehensive",
         "students_upper_secondary",
-        "students_higher_education"
+        "students_higher_education",
+        "gen_marita",
+        "gen_kalevi",
+        "att_marita",
+        "att_kalevi"
```
zptro, line 40: Could you add these to data in 08. Hankkeessa tuotetut lähtötiedot / Model-system lähtödata before merging? Otherwise model runs will fail.

Scripts/utils/freight_costs.py
```diff
@@ -0,0 +1,130 @@
...
+    Returns
+    ----------
+    road_cost : numpy.ndarray
+        impedance type cost : numpy 2d matrix
+    """
+    mode_cost = {}
+    for mode, params in unit_costs["truck"].items():
+        road_cost = (impedance["time"]*params["time_cost"]
+                     + impedance["dist"]*params["dist_cost"]
+                     + impedance["toll"]) / params["avg_load"]
```
zptro, line 27: What I meant was something like this:
```suggestion
        road_cost = sum(impedance[k] * params[k] for for k in impedance) / params["avg_load"]
```
But that means that you would have to rename the parameters "time", "dist" and "toll". And toll parameter would obviously be 1.

Scripts/utils/freight_costs.py
```diff
@@ -0,0 +1,130 @@
...
+    ship_cost = mode_cost["other_dry_cargo"]["4m"]
+    ship_aux_cost = get_aux_cost(unit_costs,
+                                 impedance,
+                                 impedance_aux)
+    return ship_cost + ship_aux_cost
+
+def get_aux_cost(unit_costs: Dict[str, Dict],
+                 impedance: Dict[str, numpy.ndarray],
+                 impedance_aux: Dict[str, numpy.ndarray]):
+    """Returns freight road auxiliary costs.
```
zptro, line 109: Could you please describe what this function does in just a bit more detail? So that it is clear that it returns `Inf` for some cases.

zptro (changes requested):
Very good! Just some small remarks.

Scripts/utils/freight_costs.py
Konnde: Right, we can also use comprehension style here as it would could simplify the func in case of road and rail calc. Not necessarily with ship calc. Can make a revision of this suggestion.

Scripts/utils/freight_costs.py
Konnde: Will update docstring here. Infs will certainly emerge, e.g. Helsinki-Espoo already being one connection pair.

Konnde:
Updated branch main zonedatas and costs with two freight commodities. Addressed PR remarks.

zptro (dismissed):

zptro (approved):

PR #100, opened 2024-11-04
Add in-vehicle time handling and possibility to calc elasticities

zptro:
I am not very excited about adding handling of these extra matrices just in case someone would like to do these elasticity calculations. Maybe we could keep the logic, but remove "unweighted_time" from `impedance_output` by default?


attesn (approved):
Seems good! I'd think time matrices are important also as network validation tool and as model result itself, not only in elasticity calculations. But admittedly not needed in many cases.

EliasMarttunen (approved):

EliasMarttunen (approved):

EliasMarttunen (approved):

PR #101, opened 2024-11-04
Remove temporary workaround

zptro:
Now when pytables is updated in the supported version of EMME, we can remove this workaround. Besides making the code cleaner, it enables error messages to tell *where* it expected to find a matrix, not only the name of the matrix it expected to find.


samakinen (approved):
Looks good

PR #102, opened 2024-11-04
Add any number of paths as possible --stored-speed-assignment arguments

zptro:
Stored speeds will be imported from these paths if set. Argument --stored-speed-assignment can also still be used as a flag, only invoking stored-speed assignment, but without importing anything.


zptro:
This now works and can be run through the UI.

juliuskroetzl (approved):

PR #104, opened 2024-11-06
Add healthcare workers to attraction variables

attesn:
Add healthcare workers to model as attaction size variable. Survey data has these trips included in location type "Place for personal business or errand (bank, office, doctor, post office, hairdresser etc.)", which are currently part of other shopping tours together with trips to "Shopping center" and "Other place for shopping".

Could not find good land-use statistics for other services (banks, offices, hairdresser), therefore those are not added.


zptro (approved):
What about other personal services? Are they part of "shop"?

attesn:
> What about other personal services? Are they part of "shop"?

Yep, there are no good statistics of other personal services in TOL 2008 classes. However, I would assume that they are located nearby shops.

PR #106, opened 2024-11-08
Impedance and demand shares with evening hour

jmutikainen:


attesn (approved):

Scripts/datatypes/purpose.py
```diff
@@ -113,15 +113,23 @@ def transform_impedance(self, impedance):
...
                         imp = impedance[time_period][mtx_type][ass_class]
+                        share = self.impedance_share[mode][time_period]
+                        outward_sum += share[0]
+                        return_sum += share[1]
                         day_imp[mode][mtx_type] += share[0] * imp[rows, cols]
                         day_imp[mode][mtx_type] += share[1] * imp[cols, rows].T
+            if abs(outward_sum/len(impedance[time_period]) - 1) > 0.001:
+                raise ValueError(f"False outward impedance shares: {self.name} : {mode}")
+            if abs(return_sum/len(impedance[time_period]) - 1) > 0.001:
+                raise ValueError(f"False return impedance shares: {self.name} : {mode}")
```
zptro, line 132: - We cannot divide the sum by `impedance[time_period]`, because the mode does not always have all impedance types. We could divide by `day_imp[mode]` instead.
- In Veeti's model, we only have outward impedance, that we want to sum to 2. So I think we could have just one comparison, checking that outward and return combined sums to 2.

zptro (changes requested):

zptro (approved):

PR #107, opened 2024-11-12
Add printing of assignment results to gpkg

zptro:
Use fiona to create gpkg file and store link and node extra attribute data in it.

### DIFF
--- a/Scripts/assignment/emme_assignment.py
+++ b/Scripts/assignment/emme_assignment.py
@@ -376,6 +376,38 @@ class EmmeAssignmentModel(AssignmentModel):
                         miles["dist"][mode] += departures * segment.link.length
                         miles["time"][mode] += departures * segment[time_attr]
         resultdata.print_data(miles, "transit_kms.txt")
+        network = self.day_scenario.get_network()
+        attr_names = [attr.name for attr in self.day_scenario.extra_attributes()
+            if attr.type == "LINK"]
+        link_recs = ({
+            "geometry": {
+                "type": "LineString",
+                "coordinates": [
+                    (link.i_node.x, link.i_node.y),
+                    (link.j_node.x, link.j_node.y),
+                ]
+            },
+            "properties": {attr: float(link[attr]) for attr in attr_names}
+        } for link in network.links())
+        schema = {
+            "geometry": "LineString",
+            "properties": {attr: "float" for attr in attr_names}
+        }
+        resultdata.print_gpkg(link_recs, "assignment.gpkg", schema, "links")
+        attr_names = [attr.name for attr in self.day_scenario.extra_attributes()
+            if attr.type == "NODE"]
+        node_recs = ({
+            "geometry": {
+                "type": "Point",
+                "coordinates": (node.x, node.y)
+            },
+            "properties": {attr: float(node[attr]) for attr in attr_names}
+        } for node in network.nodes())
+        schema = {
+            "geometry": "Point",
+            "properties": {attr: "float" for attr in attr_names}
+        }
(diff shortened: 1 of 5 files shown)


Scripts/assignment/emme_assignment.py
```diff
@@ -376,6 +376,38 @@ def aggregate_results(self,
                         miles["dist"][mode] += departures * segment.link.length
                         miles["time"][mode] += departures * segment[time_attr]
         resultdata.print_data(miles, "transit_kms.txt")
+        network = self.day_scenario.get_network()
```
samakinen, line 379: I think this should be in an own function instead of extending the already too long aggregate_results() method

Scripts/assignment/emme_assignment.py
```diff
@@ -376,6 +376,38 @@ def aggregate_results(self,
...
+        attr_names = [attr.name for attr in self.day_scenario.extra_attributes()
+            if attr.type == "LINK"]
+        link_recs = ({
+            "geometry": {
+                "type": "LineString",
```
samakinen, line 384: This will loose the actual geometry of the link. We should be able to simply use LineString(link.shape) from shapely.geometry to get the required geometry instead of constructing it our self from the node coordinates.

Scripts/datahandling/resultdata.py
```diff
@@ -119,3 +126,17 @@ def print_matrices(self,
...
+                   filename: str,
+                   schema: dict,
+                   layer: str):
+        """Save data to layer in GeoPackage file in ETRS89/GK25FIN projection.
+
+        See fiona documentation on format of records and schema.
+        """
+        with fiona.open(
+                self.path / filename, 'w', driver="GPKG", layer=layer,
+                crs=CRS.from_epsg(3879), schema=schema) as colxn:
```
samakinen, line 141: The CRS should be EPSG:3067

samakinen (changes requested):

Scripts/utils/print_links.py
```diff
@@ -1,6 +1,57 @@
...
+
+
+class LINK:
+    geom_type = "LineString"
+
+    def __new__(cls, link):
+        return LineString(link.shape)
+
+
+def geometries(scenario,
```
samakinen, line 21: scenario is of type indor.emme.database.scenario.Scenario that is dependent on Emme being available. I think this should be a Protocol (that is implemented by the Emme and Mock classes) or the function should be moved to the "emme only" parts of the code.

Scripts/utils/print_links.py
```diff
@@ -1,6 +1,57 @@
...
+
+    Returns
+    -------
+    Iterable
+        Iterator of GeoJSON records
+    dict
+        Fiona schema of record types
+    """
+    attr_names = [attr.name for attr in scenario.extra_attributes()
+        if attr.type == geom_type.__name__]
```
samakinen, line 43: This could be a separate data member in the class instead of relying on class name to have a functional purpose.

samakinen (dismissed):
Looks good. I added few comments that could be addressed or not.

Scripts/utils/print_links.py
```diff
@@ -1,6 +1,64 @@
...
+    name = "LINK"
+    geom_type = "LineString"
+
+    def __new__(cls, link):
+        return LineString(link.shape)
+
+
+def geometries(attr_names: Iterable[str],
+               objects: Iterable,
+               geom_type: GeometryType) -> Tuple[Iterable, dict]:
```
samakinen, line 32: The geom_type argument is a class and not an instance. The type annotation should be Type[GeometryType] (or type[GeometryType] for newer python)

samakinen (approved):
This looks good.

PR #112, opened 2024-11-14
Scale upper level parameters with nest coefficients

attesn:
Estimation specifies components of utility function at lowest level of mode-destination hierarchy, whereas the model-system calculates mode specific variables as utility of nest. As Alogit estimation does not perform normalization (RU1 estimation) and ensure same scale between levels, it must be done when transferring parameters from estimation to model-system.


attesn:
@zptro I'll still add re-estimated parameters for `dest<mode` models in afternoon, but code can be reviewed already.

Scripts/datatypes/purpose.py
```diff
@@ -527,14 +528,41 @@ def attempt_calibration(spec: dict):
...
+    except AttributeError:
+        # No calibration parameters in this branch, return up
+        return
+    for param_name in param_names:
+        if param_name == "scaling":
+            scale(spec, spec.pop(param_name))
+        else:
+            # Search deeper
+            attempt_scaling(spec[param_name])
+
```
zptro, line 550: Maybe we could skip this separate `attempt_scaling` method and insert this in `attempt_calibration`?
```
        elif param_name == "scaling":
            scale(spec, spec.pop(param_name))
```

zptro:
This seems to be the most clear and concise way to add the re-scaling.

zptro (dismissed):
Nice!

zptro (approved):

PR #113, opened 2024-11-15
Update number of segments in test bank

zptro:
Fixing test related to #111


zptro:
I had not updated my emmebank file.

PR #116, opened 2024-11-21
Add ferry transit line to test network

Konnde:
Added new test network transit line for mode 'w'.


Scripts/tests/emme_only/test_assignment.py
```diff
@@ -73,18 +73,18 @@ def __init__(self):
...
         self.ass_model = ass.EmmeAssignmentModel(emme_context, scenario_num)
-        dist_cost = {
+        self.dist_cost = {
             "car_work": 0.12,
             "car_leisure": 0.12,
             "trailer_truck": 0.5,
             "semi_trailer": 0.4,
             "truck": 0.3,
             "van": 0.2,
         }
```
Konnde, line 83: Changed dist cost to instance specific so that we can use it in both trip and freight tests.

Scripts/tests/emme_only/test_assignment.py
```diff
@@ -135,41 +135,17 @@ def test_assignment(self):
...
-                "firstb_single": 0.0,
-                "dist_single": 0.1
-            }
-        })
+        firstb_single = (2, 3, 5, 70, 0, 1.5)
+        dist_single = (0.1, 0.2, 0.1, 0.3, 0.1, 0.2)
+        fares = pandas.DataFrame(
+            {i: {"firstb_single": firstb_single[i],
+                 "dist_single": dist_single[i]}
+             for i in range(0, len(firstb_single))})
```
Konnde, line 143: Suggesting this slightly cleaner look on creating fare df. Made same change to test assignment in unit tests.

Scripts/tests/test_data/Network/extra_links_test.txt
```diff
@@ -4695,6 +4695,7 @@ end extra_attributes
   249548   249547         0         0
   249548   249553         0         0
   249548   393031         0         0
+  249548   900301         0         0
```
zptro, line 4698: Maybe we could add some ferry wait time to check that it works as intended?

Scripts/tests/test_data/Network/netfield_links_test.txt
zptro, line 1: I suggest we remove all records from this file and leave only the header. Car time should not even be -1, that will cause an error in STORED_SPEED_ASSIGNMENT. The default value 0 is interpreted as free-flow speed. This has recently been fixed also in network-builder.

zptro (changes requested):
Looks very good! Just some minor comments.

Konnde:
Addressed comments by adding some minutes to ferry_wait_time extra attribute and deleting netfield link records.

zptro (approved):

PR #118, opened 2024-11-26
Update tour generation to account for whether tours are open or closed

jmutikainen:
- Tour rates by tour type and age group take into account whether tour are open or closed
- Open tours are considered half of closed tours in calculations


jmutikainen:
The name of youngest age group doesn't match its data (Helmet history), which should probably be changed sometime...

zptro (changes requested):
This causes `AssertionError: 0.34241463698614005 != 0.34376735936254194 within 7 places (0.0013527223764018825 difference)` in https://github.com/Traficom/lem-model-system/blob/9ea63dbe95982967c686030ee1a2582576c5812c/Scripts/tests/integration/test_models.py#L57.

This check might seem silly, but it is to make sure all changes that cause model result changes are intentional. Could you please change the comparison value in the test?

zptro (approved):

PR #119, opened 2024-11-26
Update costdata.json

zptro:
This file is useful when testing model runs in production mode.


Konnde (approved):
Nice, more comprehensive than current iteration. Easily allows restructuring of transit cost calc.

PR #120, opened 2024-11-27
Implement first draft of freight lem execution sequence

Konnde:
First iteration implementation to run freight forecasting in lem model system environment. Draft uses similar user argument parsing to lem.py.  

Currently performs demand calculation and traversal data handling for each freight parameter given in params path. User can specify eligible commodity names that are also assigned to network. Commodity extra attribute creation is only performed for specified commodities.

This PR is currently open for discussion.


Scripts/datahandling/traversaldata.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Returns
+    ----------
+    numpy matrix
+        Aggregated auxiliary demand of freight modes
+    """
+    aux_tons = numpy.zeros([len(zones), len(zones)], dtype=numpy.float32)
+    for ass_class in param.freight_modes:
+        file = result_path / f"{ass_class}.txt"
+        aux_tons += read_traversal_file(file, numpy.array(zones))
```
Konnde, line 25: I was also thinking of passing the zeros matrix to the next func and performing addition there but this seems better to me.

Scripts/assignment/freight_assignment.py
```diff
@@ -44,12 +44,26 @@ def save_network_volumes(self, commodity_class: str):
...
             attr_name = (commodity_class + ass_class)[:17]
-            spec["on_segments"]["transit_volumes"] = '@' + attr_name
-            spec["on_links"]["aux_transit_volumes"] = '@a_' + attr_name
+            spec["on_segments"]["transit_volumes"] = "@" + attr_name
+            spec["on_links"]["aux_transit_volumes"] = "@a_" + attr_name
             self.emme_project.network_results(
                 spec, self.emme_scenario, ass_class)
+        spec = self._car_spec.truck_spec()
+        attr_name = (commodity_class + "truck")[:17]
+        spec["classes"][0]["results"]["link_volumes"] = "@" + attr_name
```
Konnde, line 53: Commodity specific truck assignment was missing. Added into this development phase.

Scripts/assignment/emme_assignment.py
```diff
@@ -183,6 +183,11 @@ def prepare_freight_network(self, car_dist_unit_cost: Dict[str, float],
                     "LINK", '@a_' + attr_name,
                     "commodity flow", overwrite=True,
                     scenario=self.mod_scenario)
+                attr_name = (comm_class + "truck")[:17]
+                self.emme_project.create_extra_attribute(
+                        "LINK", '@' + attr_name,
```
Konnde, line 188: Commodity specific truck assignment was missing. Added into this development phase.

Scripts/assignment/datatypes/freight_specification.py
```diff
@@ -69,7 +69,7 @@ def __init__(self,
                 "perception_factor": 1,
             },
             "aux_transit_time": {
-                "perception_factor": 10,
+                "perception_factor": 30,
```
Konnde, line 72: This is related to having difficulties interpreting aux time matrix results in current iteration due to timau * 3 in mode u. Instead I tested demand calc with having timau * 1 and perception factor 30 here which seemed to produce same results with easier aux time interpretation. We can drop this from this development section if needed.

Scripts/freight_lem.py
```diff
@@ -0,0 +1,194 @@
...
+        "freight_train_aux": {
+            "dist": impedance["aux_dist"]["freight_train"],
+            "time": impedance["aux_time"]["freight_train"],
+            "toll": zeros
+        },
+        "ship_aux": {
+            "dist": impedance["aux_dist"]["ship"],
+            "time": impedance["aux_time"]["ship"],
+            "toll": zeros
+        }
```
zptro, line 76: I suggest integrating `impedance_aux` into `impedance` in `calc_rail_cost`, `calc_ship_cost` and `get_aux_cost`, so that names of "aux_dist" and "aux_time" would not change. The name changes "aux_dist" -> "dist" and "aux_time" -> "time" could happen inside `get_aux_cost`. In that way you would not have to do this hard-coded remapping here. This would then be enough here:

```
impedance = {mode: {mtx_type: impedance[mtx_type][mode] for mtx_type in impedance
    if mode in impedance[mtx_type]} for mode in ("truck", "freight_train", "ship")}
```

zptro (changes requested):
Looks very good! Some remarks:

Scripts/datahandling/traversaldata.py
```diff
@@ -0,0 +1,93 @@
...
+    traversal_matrix = numpy.zeros([len(zones), len(zones)], dtype=numpy.float32)
+    with open(file) as f:
+        lines = f.readlines()
+        for line in lines:
+            line = line.strip()
+            try:
+                int(line.split(" ")[0])
+            except ValueError:
+                continue
+            filt_data = list(filter(None, line.split(" ")))
```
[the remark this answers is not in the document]
Konnde: Seems that I was stuck on splitting with white spaces and forgot that split without sep specification gets rid of them. Will update.

Scripts/freight_lem.py
```diff
@@ -0,0 +1,194 @@
...
+BASE_FOLDER = Path(__file__).parent
+BASE_ZONEDATA_FILE = "freight_zonedata.gpkg"
+
+
+def main(args):
+    log.info("Starting simulation.")
+    base_zonedata_path = Path(BASE_FOLDER, args.baseline_data_path, BASE_ZONEDATA_FILE)
+    cost_data_path = Path(BASE_FOLDER, args.cost_data_path)
+    results_path = Path(BASE_FOLDER, args.results_path)
+    emme_project_path = Path(BASE_FOLDER, args.emme_path)
```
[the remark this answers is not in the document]
Konnde: This makes sense, will update.

Scripts/freight_lem.py
```diff
@@ -0,0 +1,161 @@
...
+            matrix_counter += 7 # Shift to next 10 digit
+        ass_model.freight_network.output_traversal_matrix(resultdata.path)
+        aux_demand = transform_traversal_data(resultdata.path, zone_numbers)
+    log.info("Simulation ready.")
+
+if __name__ == "__main__":
+    parser = ArgumentParser(epilog="Freight lem-model-system entry point script.")
+    config = utils.config.read_from_file()
+    config["SAVE_EMME_MATRICES"] = True
+    config["SPECIFY_COMMODITY_NAMES"] = ["marita"]
```
zptro, line 98: If you really want to have this as the default, could you move it to `dev-config.json`?

Scripts/freight_lem.py
```diff
@@ -0,0 +1,161 @@
...
+
+    parser.set_defaults(
+        **{key.lower(): val for key, val in config.items()})
+    args = parser.parse_args()
+    
+    if not isinstance(args.specify_commodity_names, list):
+        raise TypeError("Invalid type for save commodity names")
+    for name in args.specify_commodity_names:
+        if name not in list(commodity_conversion):
+            raise ValueError("Invalid given commodity name.")
```
zptro, line 151: I think that what you want to use is the `nargs="*"` and `choices=commodity_conversion` arguments in `add_argument` for "--specify-commodity-names". Setting `type` is unnecessary when dealing with strings. See docs here: https://docs.python.org/3/library/argparse.html#nargs

Scripts/utils/freight_costs.py
```diff
@@ -113,17 +102,27 @@ def get_aux_cost(unit_costs: Dict[str, Dict],
...
+    del impedance[f"aux_dist"]
+    try:
+        impedance["time"] = impedance[f"aux_time"]
+        del impedance[f"aux_time"]
+    except KeyError:
+        pass
+    try:
+        del impedance["channel"]
+    except KeyError:
+        pass
```
zptro, line 122: Modifying an existing dict can have unpredictable effects outside the function. I suggest creating a new dict:
```suggestion
    impedance_aux = {
        "dist": impedance["aux_dist"],
        "time": impedance["aux_time"],
    }
```

Why would "aux_time" not exist?

zptro (changes requested):

Konnde:
Good comments, addressed them to improve coherency.

A matter which I noticed is that most of the current freight json parameters currently have wrong header naming. This causes the program to pick up a purpose "freight" which obviously does not exist. I have already fixed this model naming issue on the module that creates these json files. My suggestion is that, as we know that the freight parameter files in the repo are outdated due to the recently found size issue, we'll not include the updated models into this PR. Instead I will add open a new PR where we go over these updated freight models.

zptro (approved):
Looks very good!

PR #123, opened 2024-12-02
Refactor `AssignmentModel`

zptro:
- New base class `AssignmentMode` with subclasses:
  - `WalkMode` (STANDARD_TRANSIT_ASSIGNMENT specification with only aux mode)
  - `BikeMode` (STANDARD_TRAFFIC_ASSIGNMENT specification without congestion)
  - `CarMode` (part of SOLA_TRAFFIC_ASSIGNMENT specification)
  - `TruckMode` (part of SOLA_TRAFFIC_ASSIGNMENT specification)
  - `FreightMode` (special EXTENDED_TRANSIT_ASSIGNMENT specification for freight train and ship)
  - `TransitMode` (EXTENDED_TRANSIT_ASSIGNMENT specification)
- New container class `EmmeMatrix`
- Handling of "toll_cost" as a separate matrix
  - "cost" will always represent total cost (including car usage marginal cost)

This will break the matrix creation functionality in `freigth_lem.py`, but I will fix that in a subsequent PR.


Konnde (approved):

Konnde (approved):
Ok so based on code skimming, no section seems to raise a big question mark. I can notice structural updates towards new Emme api functionalities which is good. Separating assignment class specific variables into their own class instances enables not needing to hard code ass class specific variable matrices in parameters/assignment.py. The structural changes also seem to better connect assignment class and emme matrix handling which stores matrices in this "semi-stored" space where they can be then either released or dropped depending on needs.

There were some sections which might be slightly difficult to interpret, e.g. assign of freightassignmentperiod having quite deeply nested impedances. Otherwise I didn't seem to catch anything critical.

Konnde (dismissed):
Looks good, maybe slight code duplication among create emme project and lem validate inputfiles based on latest commit. Creating the nr_attributes already in parameters/assignment would solve this but I don't know if it would really improve readability.

PR #124, opened 2024-12-03
Fix distance boundaries in `DestModeModel`

zptro:
In `DestModeModel`, the lower level is mode choice, so distance boundaries must be applied when calculating mode utilities (OD-pairwise). When adding this, I discovered that freight distance boundaries had wrong name for "freight_train", so freight demand calculation did not work without name fix.


Konnde (dismissed):
Good catch.

attesn (approved):

PR #127, opened 2024-12-10
Update freight models

Konnde:
Update commodity specific freight models based on latest estimations. Added rest of the commodities to zone variables json.


Konnde:
Just an acknowledgement here; it seems that emme only assignment testing is currently failing due to the emme 23 back propagation problem caused by some developers still having old emme version.

Scripts/parameters/freight/metsat.json
```diff
@@ -26,20 +25,21 @@
...
-            "constant": 0.0,
+            "constant": 0,
             "generation": {},
             "attraction": {},
             "impedance": {
-                "cost": -0.7224
+                "cost_log": -6.524,
+                "cost": -0.2782
             },
             "log": {},
```
zptro, line 44: We have the separate "log" structure for this, where you can put "cost". It will then logarithmize it and then apply the parameter.

zptro (changes requested):
Looks good! But there is actually no need to add logic for calculating log.

Scripts/parameters/freight/metsat.json
Konnde: I see, that's what the log section is for. I'll update this particular model.

zptro (approved):

zptro:
> Simple note here; it seems that emme only assignment testing is currently failing due to the emme 23 back propagation problem caused by some developers still having old emme version.

@Konnde You mean it is failing even after #126? I was not able to reproduce any error with Emme 24. Could you specify what kind of problems you have run into?

Konnde:
@zptro  Here is an error code snippet of the encountered issue. I made a presumption that it's based on that PR due to what's being shown in the traceback. This has been run on emme 23.

geometry_error

PR #129, opened 2024-12-11
Primary destination based tours

attesn:
This PR converts non-home based full tours to primary destination based tours. Primary destination based tours are a series of linked trips starting and finishing at the primary destination of ongoing home-based tour. For example if an individual makes a lunchtime trip to the shops (and back to work) during their work day.

Using this definition makes it possible to use mode (or other data) of  'parent' tour as explanatory variable. For example arriving as car driver increases possibility to use car in pd-based tour.

Not sure what would be most user-friendly way of coding the parameters. Also we'll need to decide what is the name of 'parent' tours, so variable names are coherent in tour building, estimation and model-system.


attesn:
@zptro

zptro:
Implemented a working solution, which skips the dummy in the first demand model iteration. Need to add some logic if we want to skip the non-home-based tours entirely in the first iteration.

TODO: Update agent model

attesn:
Model run works now ok. Demand parameters require update before merge.

zptro:
Now I figured out how to calculate the mode shares correctly in the first iteration without extensive changes in LOS matrix handling. The solution is maybe not the most beautiful one.

zptro (approved):

PR #130, opened 2024-12-16
Add freight vehicle calculation

Konnde:


Konnde:
First revision of freight ton-to-vehicle conversion. As current freight lem framework only considers domestic trips, semi-trailer are omitted until further updates. Opted for vehicle_conversion.py parameter file over *.json file that would be given in baseline data path.

Scripts/freight_lem.py
```diff
@@ -86,6 +90,15 @@ def main(args):
             matrix_counter += 7 # Shift to next 10 digit
         ass_model.freight_network.output_traversal_matrix(resultdata.path)
         aux_demand = transform_traversal_data(resultdata.path, zone_numbers)
```
Konnde, line 92: alternatively we could simply add aux demand into demand["truck"] on return, and then use that matrix in calc_vehicles().

Scripts/parameters/vehicle_conversion.py
```diff
@@ -0,0 +1,207 @@
+# Variables for converting tons to vehicles for each commodity class
+
+vehicle_conversion = {
+  "marita": {
+    "truck_share": 0.059,
+    "trailer_truck_share": 0.941,
+    "truck_avg_load": 6.35,
+    "trailer_truck_avg_load": 33.32,
+    "empty_share": 0.5
```
zptro, line 9: We have the same kind of stuff in the cost data input file. Could we use that or are they different in some way?

Scripts/datatypes/purpose.py
```diff
@@ -583,3 +583,26 @@ def calc_traffic(self, impedance: dict, purpose_key: str):
...
+        args : dict
+            conversion variable : float
+
+        Returns
+        -------
+        numpy.ndarray
+            assignment class specific vehicle matrix
+        """
+        vehicles = (matrix * args[f"{ass_class}_share"] 
+                    / args[f"{ass_class}_avg_load"] * 1.02 / 365)
```
zptro, line 606: What does 1.02 represent?

Scripts/datatypes/purpose.py
```diff
@@ -583,3 +583,26 @@ def calc_traffic(self, impedance: dict, purpose_key: str):
...
+        conversion values.
+
+        Parameters
+        ----------
+        matrix : numpy.ndarray
+            ton matrix
+        ass_class : str
+            truck assignment class
+        args : dict
+            conversion variable : float
```
zptro, line 598: This seems to be purpose-specific, so I do not see the point in feeding it from the outside here.

Scripts/parameters/vehicle_conversion.py
Konnde: Hmm, we actually could fully utilize purpose specific costdatas here. For empty returning vehicle trips we could either add a new variable to the json or simply state that empty share in vehicle calc is empty share of cost calc - 1. 

The only uncertainty I have here is if the vehicle conversion ends up needing new parameters which would require to be then added into costdata json. Then that json would have variables that are not at all cost related.

Scripts/datatypes/purpose.py
Konnde: This is an addon factor which has been explained to me to signify uncertainty in the function. I wasn't really content with this approach either because it's hard coded. I think I'm going to first raise a question to the others on the necessity of this one.

Scripts/freight_lem.py
```diff
@@ -86,6 +90,15 @@ def main(args):
...
+        truck_demand = demand["truck"] + aux_demand
+        for mode in ("truck", "trailer_truck"):
+            total_demand[mode] += purpose_value.calc_vehicles(truck_demand,
+                                                              mode,
+                                                              vehicle_conversion[purpose])
+    for ass_class in total_demand:
+        ass_model.freight_network.set_matrix(ass_class, total_demand[ass_class])
+    ass_model.emme_project.car_assignment(ass_model.freight_network._car_spec._spec,
+                                          ass_model.freight_network.emme_scenario)
```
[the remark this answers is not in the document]
Konnde: Yes, way better.

Scripts/parameters/vehicle_conversion.py
zptro: I guess you could leave the non-cost related parameters in this file.

zptro (approved):
Looks very good!

PR #131, opened 2024-12-16
Update demand parameters with fixed impedance

attesn:
Return trip costs calculated correctly. Only small changes to mode shares.


zptro (changes requested):
Just fix this and we are good! `AssertionError: 0.33648806471355974 != 0.335967979242067`

zptro (approved):

PR #132, opened 2024-12-18
Refactor freight model slightly

zptro:
Move cost calculation inside purpose and make mode-switch. Utilize freight purpose name from specification.

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -15,6 +15,7 @@ import parameters.cost as cost
 import models.generation as generation
 from datatypes.demand import Demand
 from datatypes.histogram import TourLengthHistogram
+from utils.freight_costs import calc_cost
 
 
 class Purpose:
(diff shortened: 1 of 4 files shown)


zptro:
> Nice refactoring! Better utilization of purpose name in contrast to current iteration where it is passed in explicitly.

The `match/case` solution requires Python 3.10 ->, but is this otherwise ready to merge?

Konnde:
> > Nice refactoring! Better utilization of purpose name in contrast to current iteration where it is passed in explicitly.
> 
> The `match/case` solution requires Python 3.10 ->, but is this otherwise ready to merge?

Yes, I have a working 3.7 copy so we can proceed with this one. Then, hopefully at the beginning of the year our enterprise releases us those new Emme licenses, enabling dev continuation here.

zptro:
> > > Nice refactoring! Better utilization of purpose name in contrast to current iteration where it is passed in explicitly.
> > 
> > 
> > The `match/case` solution requires Python 3.10 ->, but is this otherwise ready to merge?
> 
> Yes, I have a working 3.7 copy so we can proceed with this one. Then, hopefully at the beginning of the year our enterprise releases us those new Emme licenses, enabling dev continuation here.

So... approve? 😅

Konnde (changes requested):
Looking good. Added one note.

Konnde (approved):

PR #134, opened 2024-12-31
Speed up 24h expansion and bee-line calculation

zptro:
Speeds up 24h expansion for whole Finland from two hours to three minutes. Beeline calculation does not really matter (takes a few seconds anyway), but new solution maybe a bit more elegant.


samakinen (dismissed):
Very nice fix. Looks good.

PR #135, opened 2025-01-10
Enable network acceleration in car assignment

zptro:


zptro:
In last test with new enlarged and optimized network (connections abroad), Uusimaa aht car assignment in loop goes from 9 min to 6,5 min.

PR #138, opened 2025-01-14
Create new `WholeDayPeriod` for long-dist assignment

zptro:
Add separate EMME assignment definition for long-distance trips. This period represents the whole day and only long-distance modes. Cars are assigned with free-flow speed. `AssignmentPeriod` init parameter `use_free_flow_speeds` is dropped, because it is now set explicitly in `WholeDayPeriod` and `FreightPeriod`. Also added object variable `transport_classes`, which represents the main assignment classes of the model run.

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -24,55 +24,49 @@ class AssignmentPeriod(Period):
     This typically represents an hour of the day, which may or may not
     have a dedicated EMME scenario. In case it does not have its own
     EMME scenario, assignment results are stored only in extra attributes.
-
-    Parameters
-    ----------
-    name : str
-        Time period name (aht/pt/iht)
-    emme_scenario : int
-        EMME scenario linked to the time period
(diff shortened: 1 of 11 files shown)


Konnde (dismissed):
The structure seems to look good based on brief screening. Off_peak_period.py seems to have invalid import clause for numpy ndarray as it tries to import it through core? Same file also seems to have unused WholeDayperiod import clause.

zptro:
> The structure seems to look good based on brief screening. Off_peak_period.py seems to have invalid import clause for numpy ndarray as it tries to import it through core? Same file also seems to have unused WholeDayperiod import clause.

Good catch with the `ndarray` import! Now fixed. `WholeDayPeriod` is imported, because it is used here: https://github.com/Traficom/lem-model-system/blob/refactor/long-distance-assignment/Scripts/assignment/emme_assignment.py#L110. Maybe not the most clear way of defining the assignment periods though.

zptro:
According to bad practice, I added some minor fixes related to #123 to this PR. Walk mode and toll costs were not handled correctly, which would have caused issues later on.

PR #140, opened 2025-01-17
Do zone-data variable multiplication on the fly

zptro:
The original point of this PR was to create "within_zone_dist" and "within_zone_time" matrices on the fly, because I did not want them to take up memory permanently. But then I realized that what we want to do is to multiply two variables in general, so the magic sign should be `*`. From now on, we can use it for multiplication of any two variables.


attesn (approved):
Excellent! I'll change the templates in estimation repository.

PR #143, opened 2025-01-24
Print long distance transit boardings

EliasMarttunen:
Print boardings only for long distance transit if free_flow_assignment is used


zptro (approved):
Looks good!

PR #148, opened 2025-01-29
Electric car handling

zptro:
- Added `bev` and `phev` as own assignment classes with lower distance cost and same value of time as `icev` (aka `car`).
- Added county-wise "car_shares" parameter to cost-data file, model-system then creates a zone data variable from it.
- The demand model performs mode choice calculation three times per tour purpose, one for traditional car owners (mode `car`) and one for each electric car mode. Parameter "car_shares" is used for calculating the weighted average mode choices (with the three car types as separate modes).
- Logsum calculations are still performed only for the traditional cars. The choice between electric and traditional cars is not a discrete choice model (yet), so a super-logsum cannot easily be constructed.


zptro:
After adding better "car_pax" handling, this branch works, but still needs some code improvements. To be able to run it with the whole Finland setup, #160 must first be merged.

attesn:
> * The demand model performs mode choice calculation twice per tour purpose, one for traditional car owners (modes `car_work` and `car_leisure`) and one for electric cars. Parameter "share_electric_cars" is used for calculating the weighted average mode choices (with the two car types as separate modes).
> * Logsum calculations are still performed only for the traditional cars. The choice between electric and traditional cars is not a discrete choice model (yet), so a super-logsum cannot easily be constructed.

Would it be possible to split electric cars when moving demand matrices to assignment model and return average car impedances (fuel + electric = car) to logit choice models? I find this implementation in `logit.py` to be too complicated, when there is no choice model. Also share of electric cars is input data, so there is no benefits of having them to output files.

zptro:
> > * The demand model performs mode choice calculation twice per tour purpose, one for traditional car owners (modes `car_work` and `car_leisure`) and one for electric cars. Parameter "share_electric_cars" is used for calculating the weighted average mode choices (with the two car types as separate modes).
> > * Logsum calculations are still performed only for the traditional cars. The choice between electric and traditional cars is not a discrete choice model (yet), so a super-logsum cannot easily be constructed.
> 
> Would it be possible to split electric cars when moving demand matrices to assignment model and return average car impedances (fuel + electric = car) to logit choice models? I find this implementation in `logit.py` to be too complicated, when there is no estimated choice models.

It took me 1½ years to realize that you were right about this. Not only because the logit implementation was complicated, but also because the mode choices are meaningless if not connected to the vehicle-type choice. People who drive much will increasingly buy electrical cars, so there is a strong self-reenforcing loop that need to be accounted for.

Scripts/tests/test_data/Scenario_input_data/costdata.json
```diff
@@ -12,6 +13,28 @@
         "semi_trailer": 35.11,
         "trailer_truck": 36.0
     },
+    "car_shares": {
```
attesn: Would it be more user-friendly to enter shares as new column in zonedata geopackage? There would be redundant data in column, as share is same within county, but importing one zone data variable using different syntax adds complexity from my viewpoint.  

And we'll probably end up having cases where Helsinki wants to use their own forecast of electric car shares. :)

attesn (changes requested):
Code seems good for me. One comment to discuss about importing new share data.

attesn:
> It took me 1½ years to realize that you were right about this. Not only because the logit implementation was complicated, but also because the mode choices are meaningless if not connected to the vehicle-type choice. People who drive much will increasingly buy electrical cars, so there is a strong self-reenforcing loop that need to be accounted for.

Good point, didn't think of this either. We'd probably need to discuss about how to handle electric cars in future model versions, as this might be one of the strongest effects shaping the future mode choices at national level.

Scripts/tests/test_data/Scenario_input_data/costdata.json
zptro: For me this would be fine, it would in fact simplify the code somewhat. But I am afraid this would make the data process more complicated, as @rviri1 would have to not only coordinate results from the car-fleet model with other cost calculations, but also with zone data formation. How simple would it be for you to create a data flow process where all these things would be built together?

zptro: This concerns both formation of baseline data by @attesn and of forecast data by @koskivaltteri.

rviri1: In general, this approach would not need a any major change, as zonedata geopackages are imported during fleet generation. Currently after import, multiple columns (including geometry are currently dropped) and only the ones regarding car fleet are kept and scaled. However, at that point, it is also possible to keep a copy of the original (for each snapshot year), and at the end of car cost calculation add bev_mileage_share and phev_mileage_share as new columns to the original and then output new geopackage. 

So either case works from this viewpoint, adding the shares to the geopackage might mean a bit of complesixity for the user, as running different price variations in the model would mean, that instead of replacing costs.json, also zonedata need to be replaced.

PR #151, opened 2025-01-31
Calibration v.0.1.3

EliasMarttunen:
Mode choice calibration for model v.0.1.3


zptro (approved):
I changed the base branch to main, so that dev will stay un-calibrated. A little worried about the large change in car mode share, could there be an error somewhere?

attesn:
Probably there is some issue here as calibration should not cause shift in mode shares. Calibration data has remained constant, so it should keep mode shares the same.

PR #156, opened 2025-02-07
Remove car_work and transit_work LOS matrices after usage

zptro:
This makes model runs more memory efficient. First "work"-type tour purposes are calculated, and after that "car_work" and "transit_work" LOS matrices can be dropped.


attesn (approved):
Works fine! 

This is not related to code changes, but I would consider if we need to have work and leisure assignment classes in model-system. They cause memory issues, require extra efforts when summarizing results and main benefit is more accurate CBA consumer surplus calculation. Assignment parameters are very similar for both groups. 

Could we get approximately same results by storing work and leisure split only in last iteration?

PR #159, opened 2025-02-11
Change freight lem specific flag to save commodity omx file

Konnde:
Instead of allowing user to specify which commodities are saved and assigned into network as tons, allow user to flag whether an omx-file is saved that contains commodity-mode pairs in tons.


Scripts/tests/unit/test_assignment.py
```diff
@@ -105,11 +105,10 @@ def test_long_dist_assignment(self):
 
     def test_freight_assignment(self):
         ass_model = EmmeAssignmentModel(self.context, self.scenario_id)
-        ass_model.prepare_freight_network(self.dist_cost, ["c1", "c2"])
+        ass_model.prepare_freight_network(self.dist_cost)
         ass_model.freight_network.assign()
         demand = numpy.full((ass_model.nr_zones, ass_model.nr_zones), 1.0)
         for mode in ["truck", "freight_train", "ship"]:
             ass_model.freight_network.set_matrix(mode, demand)
-        ass_model.freight_network.save_network_volumes("c1")
```
Konnde, line 113: Going through this change means that test freight assignment is unable to assign c1 freight train and ship as those extra attributes are not created anymore in prepare_freight_network. We could enforce this extra attribute creation in this test.

Scripts/freight_lem.py
```diff
@@ -65,12 +61,9 @@ def main(args):
         demand = purpose.calc_traffic(impedance)
         for mode in demand:
             ass_model.freight_network.set_matrix(mode, demand[mode])
-            if purpose.name not in args.specify_commodity_names:
-                continue
-            with resultmatrices.open("freight_demand", "vrk", zone_numbers, m="a") as mtx:
-                mtx[f"{purpose}_{mode}"] = demand[mode]
-        if purpose.name in args.specify_commodity_names:
-            ass_model.freight_network.save_network_volumes(purpose.name)
```
zptro, line 73: So if `save_network_volumes` is not used anymore, should that method be removed as well?

Scripts/freight_lem.py
```diff
@@ -65,12 +61,9 @@ def main(args):
...
+            if args.save_freight_matrix:
+                with resultmatrices.open("freight_demand", "vrk", zone_numbers, m="a") as mtx:
+                    mtx[f"{purpose.name}_{mode}"] = demand[mode]
```
zptro, line 66: What are you planning to do with this massive amount of massive matrices? Is that something that the user will want to analyze?

Scripts/freight_lem.py
Konnde: Yes I should have removed the method during committing. Depending on whether it would be beneficial/unnecessary to assign purpose during freight testing, I can either drop the method or we can can keep it just for testing.

Scripts/freight_lem.py
Konnde: That's the thing, I personally don't see purpose for it currently but others may (and already think) otherwise, as in, there are benefits in getting to observe produced ton matrices. In order to reduce matrix size for those users who desire to have only specified commodities saved, we could combine save_freight_matrix and specify_commodity_names arguments where user ticks boxes in UI to declare which commodities they want to save.

zptro: Could we instead of this new flag then simply use the old argument and save matrices only `if purpose.name in args.specify_commodity_names`?

Scripts/freight_lem.py
zptro: Well, I think that depends on what we mean by "freight testing".

If we mean validating the results coming out of the model when running it with certain input, I think that is an integral feature of the model that should be possible to do through the user interface. In that case I do not think we should remove this feature at all.

If we, on the other hand, mean verifying that the scripts produce the intended type of output, all scripts (including `save_network_volumes`) used for such testing should be in the unit tests, and not in the model system itself.

Konnde: I do agree that the method could be utilized in both of the mentioned cases with different purpose. In that case we could keep the method as it is and utilize it in both situations, during freight_lem run and test freight assignment unittest.

Considering updates, while the freight assignment unittest could be as it is. For this freight model validation feature through UI, we could pre-define one commodity for which extra attributes are created. This could be then assigned and validations for mileage etc could be performed. This would combine model validation purposes while avoiding creating unnecessary mode and commodity specific extra attributes.

PR #160, opened 2025-02-12
Separate bike and walk probability calculation

zptro:
Bike and walk utilities are calculated and stored in a separate step before actual mode choice, so that their LOS matrices can then be removed. Later on, destination choice for these modes is done separately.

This also opens up the possibility to have a denser zone system for these modes.


zptro:
After #156 has been merged, we should add popping of bike and walk LOS matrices in ModelSystem.

attesn (approved):

attesn (approved):

PR #162, opened 2025-02-18
Add car ownership choice model

attesn:
Add car ownership choice model that predicts car density for each zone. Only one individual level variable is used to enable model-system run without agent simulation.

Primary purpose here is to tests choice model specification with full population. Better integration with mode-destination choice is under development.


attesn:
@zptro It seems that there is an error in calculation of split between one and two cars. We did spreadsheet calculation that yields different results, but I just can't find the error.

zptro (dismissed):
Looks really good!

zptro (approved):

PR #164, opened 2025-02-26
Update tour generation rates

jonisylvinwsp:
Update tour generation rates for different hb age groups and update mode share accordingly.


Scripts/parameters/tour_generation.py
```diff
@@ -61,67 +61,67 @@
...
+        "age_50_64": 0.002,
         "age_65_99": 0
     },
     "hb_edu_higher": {
-        "age_7_17": 0.0,
-        "age_18_29": 0.0,
-        "age_30_49": 0.0,
-        "age_50_64": 0.0,
-        "age_65_99": 0.0
+        "age_7_17": 0.0046,
```
attesn, line 85: These could be set to zero as higher education tours already have logit tour generation model (parameters above).

attesn (changes requested):
One comment. My mistake really as I forgot to say about higher education tours.

attesn (approved):
Excellent!

PR #170, opened 2025-03-11
Enable same dummy variable for multiple modes

zptro:
Also fix bug in calculation of no-dummy shares


zptro:
We still do not have any dummy-variable handling in `DestModeModel`! I will start implementing that now.

zptro:
Done.

attesn (approved):
Seems good!

PR #174, opened 2025-03-13
Write freight specific summary txt files

Konnde:
Draft for writing freight lem specific summary txt files.

### DIFF
--- a/Scripts/freight_lem.py
+++ b/Scripts/freight_lem.py
@@ -3,6 +3,7 @@ import sys
 from pathlib import Path
 import numpy
 import json
+from pandas import DataFrame
 
 import utils.log as log
 import utils.config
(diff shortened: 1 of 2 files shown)


Scripts/freight_lem.py
```diff
@@ -50,6 +51,7 @@ def main(args):
     
     ass_model.prepare_freight_network(costdata["car_cost"], purps_to_assign)
     impedance = ass_model.freight_network.assign()
+    truck_distances = {key: impedance["dist"][key] for key in param.truck_classes}
```
Konnde, line 54: This is currently used in ```write_vehicle_summary``` to calc truck class specific mileage, as currently ```impedance dict``` contains info only for freight modes. As of this development stage, we could alternatively skip this and only use truck dist for all vehicle mileage calculations in write_vehicle_summary.

zptro (changes requested):

PR #175, opened 2025-03-13
Improve freight assignment specification

zptro:
- Move aux truck time perception factor to specification
  - This will override mode speed
- Add canal cost and enable toll cost
- Also fix performance settings


Konnde (approved):
Very nice. Now we get to validate results with Saimaa canal also being obstructed in assignment too 👍

PR #181, opened 2025-03-20
Clear long-dist matrices in off-peak period

zptro:
In off-peak periods, LOS matrices are created only once and then stored in EMME for later iterations. However, it is not necessary to store LOS matrices for long-distance modes.

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -682,7 +682,8 @@ class AssignmentPeriod(Period):
                                              / (2.0*line[effective_headway_attr]))
         self.emme_scenario.publish_network(network)
 
-    def _assign_transit(self, transit_classes=param.local_transit_classes):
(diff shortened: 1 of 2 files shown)


EliasMarttunen (commented):
Raises name error in long distance forecast if the baseline matrices folder is not found: Exception has occurred: NameError
Baseline matrix directory does not exist.
  File "C:\Users\martteli\Documents\model_system\Scripts\lem.py", line 58, in main
    raise NameError

zptro:
> Raises name error in long distance forecast if the baseline matrices folder is not found: Exception has occurred: NameError Baseline matrix directory does not exist. File "C:\Users\martteli\Documents\model_system\Scripts\lem.py", line 58, in main raise NameError

I do not think that error has anything to do with this PR.

EliasMarttunen (approved):

PR #198, opened 2025-04-08
Make demand matrix definition for assignment periods more explicit and release matrices after end assignment

zptro:
- Make demand matrix definition for assignment periods more explicit
  - `DepartureTimeModel` adds demand only for assignment periods where mode is initialized
- Remove all but one car mode from `TransitAssignmentPeriod`
- Release matrices after end assignment


zptro:
Still some small problems to solve

EliasMarttunen (approved):

PR #200, opened 2025-04-09
Remove saving of LOS matrices when MockAssignment is used

zptro:
MockAssignment gets the data from the same files where we later save them, so saving them is really unnecessary.


attesn (approved):
Excellent!

PR #202, opened 2025-04-11
Add long-dist demand calculation test and fix car ownership

zptro:


Scripts/demand/trips.py
```diff
@@ -210,7 +210,12 @@ def _get_probs(self, age: str, is_car_user: bool) -> pandas.DataFrame:
         return pandas.DataFrame(probs).to_numpy().cumsum(axis=1)
     
     def calculate_car_ownership(self, impedance):
-        acc_purpose = self.purpose_dict["hb_leisure"]
+        try:
```
attesn, line 213: Do I read correctly that car ownership is calculated only with >1 iterations? That would be ok as well, just to know.

.gitignore
```diff
@@ -62,5 +62,10 @@ Scripts/tests/test_data/Results/*
 !Scripts/tests/test_data/Results/test/Matrices/uusimaa/time*
```
attesn, line 62: I think `uusimaa` and `koko_suomi` matrices have same zones, so could have name for clarity.

attesn (approved):
Looks good! One minor tweak below.

.gitignore
zptro: When running long-distance demand calculation through the UI, `submodel` is locked to "koko_suomi". I put it like this so that small-case UI tests can be easily performed.

Scripts/demand/trips.py
zptro: No, it means that car-ownership model is run only if the demand model has a tour purpose called "hb_leisure", for which it can perform the accessibility calculation. The long-distance demand model does not have such a purpose, so then we skip the whole thing.

PR #203, opened 2025-04-11
Remove attribute requirements for non-transit lines

zptro:
This makes it possible to have freight lines in the network lacking headway attributes and fares when running transit assignment.

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -264,20 +264,22 @@ class AssignmentPeriod(Period):
         network = self.emme_scenario.get_network()
         penalty_attr = param.line_penalty_attr.replace("us", "data")
         op_attr = param.line_operator_attr.replace("ut", "data")
-        long_dist_transit_modes = {mode for mode_set
-            in param.long_dist_transit_modes.values() for mode in mode_set}
+        long_dist_transit_modes = list({mode for mode_set
+            in param.long_dist_transit_modes.values() for mode in mode_set})
+        transit_modes = long_dist_transit_modes + param.local_transit_modes
         for mode in long_dist_transit_modes:
             if network.mode(mode) is None:
                 raise AttributeError(f"Long-dist mode {mode} does not exist.")
         for line in network.transit_lines():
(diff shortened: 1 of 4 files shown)


Konnde:
I validated these with quick testing and they seem to work, great! While we're at it, should we also handle currently emerging config arg and zonedata errors that are passenger scenario related in **lem_validate_inputfiles.py**

This one is currently being raised for not passing base matrices which freight_lem does not require. We could extend current handling process and include _if not calculate_long_dist_demand or freight_demand_ etc.
https://github.com/Traficom/lem-model-system/blob/a0a438370269f0c4b877733a7a25b5b92a1a201a/Scripts/lem_validate_inputfiles.py#L184

This one is raised for missing passenger related zone data in freight_zonedata.gpkg. This one should distinguish whether given zonedata is for passenger or for freight forecasting.
https://github.com/Traficom/lem-model-system/blob/a0a438370269f0c4b877733a7a25b5b92a1a201a/Scripts/lem_validate_inputfiles.py#L197

zptro:
> I validated these with quick testing and they seem to work, great! While we're at it, should we also handle currently emerging config arg and zonedata errors that are passenger scenario related in **lem_validate_inputfiles.py**
> 
> This one is currently being raised for not passing base matrices which freight_lem does not require. We could extend current handling process and include _if not calculate_long_dist_demand or freight_demand_ etc.
> 
> https://github.com/Traficom/lem-model-system/blob/a0a438370269f0c4b877733a7a25b5b92a1a201a/Scripts/lem_validate_inputfiles.py#L184

That one is actually changing in #187 so that the condition is `if submodel != "koko_suomi"`.

> 
> This one is raised for missing passenger related zone data in freight_zonedata.gpkg. This one should distinguish whether given zonedata is for passenger or for freight forecasting.
> 
> https://github.com/Traficom/lem-model-system/blob/a0a438370269f0c4b877733a7a25b5b92a1a201a/Scripts/lem_validate_inputfiles.py#L197

Yes, I suggest you go ahead with a separate PR for that.

Konnde (approved):
Discussion points will be addressed in other PRs.

PR #206, opened 2025-04-14
Export average tour distance by mode

attesn:
Result that cannot be drived from current output data, but would be useful in when visualizing emissions and other mode-destination models results. 

Network distance would be better than beeline, but I could not figure out clever way of accessing network distances while keeping memory usage optimized.


Scripts/modelsystem.py
```diff
@@ -483,23 +501,30 @@ def _export_model_results(self):
...
-        return int_demand
+                int_dist[bounds] += purpose.generated_distance[mode]
+        return int_demand, int_dist / int_demand
     
     def _attracted_tours_mode(self, mode):
-        int_demand = pandas.Series(0, self.zdata_forecast.zone_numbers, name=mode)
+        int_demand = pandas.Series(
+            0, self.zdata_forecast.zone_numbers, name=mode)
+        int_dist = pandas.Series(
+            0.0, self.zdata_forecast.zone_numbers, name=mode)
```
zptro, line 519: Considering the way we use this method (and _attracted_tours_mode()`, I think we should change it into `_generate_tours()`, which would return to dicts with mode as key. In this way, we could reuse the dicts in `_export_model_results` instead of constructing them multiple times.

Scripts/modelsystem.py
```diff
@@ -483,23 +501,30 @@ def _export_model_results(self):
...
         for purpose in self.dm.tour_purposes:
             if mode in purpose.modes and purpose.dest != "source":
                 bounds = (next(iter(purpose.sources)).bounds
                     if isinstance(purpose, SecDestPurpose)
                     else purpose.bounds)
                 int_demand[bounds] += purpose.generated_tours[mode]
-        return int_demand
+                int_dist[bounds] += purpose.generated_distance[mode]
+        return int_demand, int_dist / int_demand
```
zptro, line 513: Considering the way we use this method (and _attracted_tours_mode()`, I think we should change it into `_generate_tours()`, which would return to dicts with mode as key. In this way, we could reuse the dicts in `_export_model_results` instead of constructing them multiple times.

zptro (approved):
Looks very good!

Scripts/modelsystem.py
```diff
@@ -480,26 +492,27 @@ def _export_model_results(self):
...
-        int_demand = pandas.Series(0, self.zdata_forecast.zone_numbers, name=mode)
-        for purpose in self.dm.tour_purposes:
-            if mode in purpose.modes and purpose.dest != "source":
-                bounds = (next(iter(purpose.sources)).bounds
-                    if isinstance(purpose, SecDestPurpose)
-                    else purpose.bounds)
-                int_demand[bounds] += purpose.attracted_tours[mode]
-        return int_demand
+    def _get_mode_tours(self, generation = True):
+        tours = defaultdict(dict)
```
zptro, line 496:
```suggestion
        tours = {"tours": {}, "dist": {}}
```

zptro: This would maybe improve readability somewhat.

PR #209, opened 2025-04-16
Make car assignment sequential when `use_stored_speeds`

zptro:
With more than 9,000 zones, adding more assignment classes seems to be causing memory issues in car assignment. Because we do not use congested assignment for this large network, we do not have to assign all classes simultaneously. This PR applies a sequential assignment process when `use_stored_speeds` is enabled.


Konnde (approved):
Great, the improved memory handling seems to be achieved with use of yielding in sequential assignment.

PR #210, opened 2025-04-17
Add linear sustainable accessibility model and subsequent mode share

jonisylvinwsp:
Add linear sustainable accessibility model and subsequent mode share change.


zptro (approved):
Looks good!

PR #215, opened 2025-04-23
Fix actual aux transit cost result spec nesting level

Konnde:
Fix for moving toll cost mf id to correct nesting level for local_result_spec in freight spec. Otherwise Emme raises type error on matrix_results func with local_result_spec


zptro (approved):
Thanks!

PR #217, opened 2025-04-23
Update freight lem parse arguments

Konnde:
Add missing parse args (forecast data path and del strat files) that are used in freight lem UI configurations. Add new trade demand parse argument for handling foreign forecasting demand. This is not yet in use. Delete deprecated cost function imports from freight lem.

### DIFF
--- a/Scripts/freight_lem.py
+++ b/Scripts/freight_lem.py
@@ -20,7 +20,7 @@ from parameters.commodity import commodity_conversion
 
 
 def main(args):
-    base_zonedata_path = Path(args.baseline_data_path, args.forecast_data_path)
+    zonedata_path = Path(args.forecast_data_path)
     cost_data_path = Path(args.cost_data_path)
     results_path = Path(args.results_path, args.scenario_name)
     emme_project_path = Path(args.emme_path)
@@ -31,7 +31,7 @@ def main(args):
                                     save_matrices=args.save_emme_matrices,
                                     first_matrix_id=args.first_matrix_id)
     zone_numbers = ass_model.zone_numbers
-    zonedata = FreightZoneData(base_zonedata_path, zone_numbers, "koko_suomi")
+    zonedata = FreightZoneData(zonedata_path, zone_numbers, "koko_suomi")
     resultdata = ResultsData(results_path)
     resultmatrices = MatrixData(results_path / "Matrices" / "koko_suomi")
     costdata = json.loads(cost_data_path.read_text("utf-8"))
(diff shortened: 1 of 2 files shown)


Scripts/freight_lem.py
```diff
@@ -16,14 +16,11 @@
...
 from datahandling.traversaldata import transform_traversal_data
-from utils.freight_costs import calc_rail_cost, calc_road_cost, calc_ship_cost
 from parameters.commodity import commodity_conversion
 
-BASE_ZONEDATA_FILE = "freight_zonedata.gpkg"
-
 
 def main(args):
-    base_zonedata_path = Path(args.baseline_data_path, BASE_ZONEDATA_FILE)
+    base_zonedata_path = Path(args.baseline_data_path, args.forecast_data_path)
```
zptro, line 23:
```suggestion
    base_zonedata_path = Path(args.forecast_data_path)
```

I do not think we need the baseline_data_path here. Could we remove that argument all together?

zptro (changes requested):
Looks good! Just one small remark:

Scripts/freight_lem.py
Konnde: Good catch, you're correct that we have no need for this argument as there are not baseline datas in freight lem runs.

Scripts/freight_lem.py
```diff
@@ -145,9 +142,9 @@ def write_vehicle_summary(demand: dict, dist: dict, resultdata: ResultsData):
         type=str,
         help="Scenario name"),
     parser.add_argument(
-        "--baseline-data-path",
+        "--forecast-data-path",
         type=str,
-        help="Path to folder containing both baseline zonedata and -matrices"),
+        help="Path to folder containing forecast zonedata"),
```
zptro, line 147:
```suggestion
        help="Path to file containing forecast zonedata"),
```

zptro (changes requested):

zptro (dismissed):

Scripts/freight_lem.py
```diff
@@ -16,25 +16,26 @@
...
+    zonedata_path = Path(args.forecast_data_path)
     cost_data_path = Path(args.cost_data_path)
     results_path = Path(args.results_path, args.scenario_name)
     emme_project_path = Path(args.emme_path)
     parameters_path = Path(__file__).parent / "parameters" / "freight"
-    
+    if args.specify_commodity_names is None:
+        log.debug("Argument --specify-commodity-names not defined."
+                  " Setting argument value as empty list.")
+        args.specify_commodity_names = []
```
Konnde, line 31: Added to prevent UI users from needing to explicitly specify the argument. Otherwise filtering purposes to assign will raise type error.

zptro: I guess the same could also be achieved by initializing the argument as `[]` here: https://github.com/Traficom/lem-model-system/blob/dev/Scripts/utils/config.py#L64

zptro: No, it should not cause any problems in lem.py, just as the attributes used only in lem.py do not cause problems in freight_lem.py.

zptro (approved):
Nice!

PR #223, opened 2025-04-29
Remove separate `AccessibilityModel`

zptro:
After geographic sub-bounds were removed in #192, the separate `AccessibilityModel` is no longer needed. I moved `calc_accessibility()` to the regular probability calculation as an optional extra calculation step.


attesn (dismissed):
Seems good!

attesn (approved):

PR #224, opened 2025-05-02
Add ut2 value for freight lines in test data and fix emme_only test

zptro:
Attribute `ut2` is used as effective headway in transit assignment. It seems that this attribute cannot be zero in transit assignment, even for modes that are not included in the assignment. Now I added a value for freight lines, just to make it work.

@Konnde will *all* freight lines have `ut2>0` in upcoming versions of the network, or do we need to come up with a workaround?

### DIFF
--- a/Scripts/tests/emme_only/test_assignment.py
+++ b/Scripts/tests/emme_only/test_assignment.py
@@ -104,7 +104,8 @@ class EmmeAssignmentTest:
         self.ass_model.init_assign()
         for ap in self.ass_model.assignment_periods:
             for ass_class in demand:
-                ap.set_matrix(ass_class, car_matrix)
+                if ass_class in ap.assignment_modes:
+                    ap.set_matrix(ass_class, car_matrix)
             travel_cost[ap.name] = ap.end_assign()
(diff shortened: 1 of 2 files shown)


Konnde:
> Attribute `ut2` is used as effective headway in transit assignment. It seems that this attribute cannot be zero in transit assignment, even for modes that are not included in the assignment. Now I added a value for freight lines, just to make it work.
> 
> @Konnde will _all_ freight lines have `ut2>0` in upcoming versions of the network, or do we need to come up with a workaround?

External transit lines (foreign ship transit lines traversing between Finnish ports and foreign ports) are planned to have ut2 data (bitbucket freight-maritime-updates PR, ut2 seems to now be frequency in current development phase, and frequency should always be larger than 0). However, domestic mode "W" currently has no needs for ut2. Would you consider this a problem if we set default ut2 value for W to be e.g. 1 for every "W" transit line link in koko suomi network?

Konnde (approved):
We will proceed with these changes for now, as in, expect that ut2 will always have value larger than 0. This might change in future updates.

PR #226, opened 2025-05-02
Set in-vehicle time weight to 0.7 for long-distance train

zptro:
We could try this to see if we can get more long transit tours. This is a much more simple approach than having a separate train mode in mode choice.

This must be accompanied by cost damping, because otherwise these tours will have lower value of time than other tours, which is contradictory to what SP studies say.

Update 2025-12-11: This can also be tested to see if we can avoid "train" tours that include too long connecting bus legs.


attesn (commented):
Code seems good and next we could run full model to produce impedance matrices for estimation. However should we check destination aggregation effects separately?

attesn:
This was merged as part of #308.

PR #228, opened 2025-05-06
Aggregate far-away destinations

zptro:
Handle destinations that are so far away that their probability would be rounded to zero, by aggregating their size variable in chunks of 50. Also remove deprecated handling of matrix transformations.

### DIFF
--- a/Scripts/models/logit.py
+++ b/Scripts/models/logit.py
@@ -80,14 +80,11 @@ class LogitModel:
 
     def _calc_dest_util(self, mode: str, impedance: dict) -> numpy.ndarray:
         b = self.dest_choice_param[mode]
-        utility: numpy.array = numpy.zeros_like(next(iter(impedance.values())))
-        self._add_zone_util(utility, b["attraction"])
-        self._add_impedance(utility, impedance, b["impedance"])
-        size = numpy.zeros_like(utility)
+        size = numpy.zeros_like(next(iter(impedance.values())))
         self._add_zone_util(size, b["attraction_size"])
         impedance["attraction_size"] = size
-        self._add_log_impedance(utility, impedance, b["log"])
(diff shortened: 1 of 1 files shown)


Scripts/models/logit.py
```diff
@@ -86,13 +86,24 @@ def _calc_dest_util(self, mode: str, impedance: dict) -> numpy.ndarray:
         size = numpy.zeros_like(utility)
         self._add_zone_util(size, b["attraction_size"])
         impedance["attraction_size"] = size
-        if "transform" in b:
-            b_transf = b["transform"]
-            transimp = numpy.zeros_like(utility)
-            self._add_zone_util(transimp, b_transf["attraction"])
-            self._add_impedance(transimp, impedance, b_transf["impedance"])
-            impedance["transform"] = transimp
```
zptro, line 94: Now we suddenly came up with a use case for this, so I will need to modify this branch a little bit to take into account "transform".

Scripts/models/logit.py
```diff
@@ -86,13 +86,24 @@ def _calc_dest_util(self, mode: str, impedance: dict) -> numpy.ndarray:
...
+        # The utility of these destinations is so small that they would
+        # get zero probability
+        if too_small.any():
+            too_small_dests = size[too_small]
+            chunk_size = 50
+            too_small_dests_padded = numpy.pad(
+                too_small_dests, (0, -too_small_dests.size % chunk_size))
+            chunks = too_small_dests_padded.reshape(-1, chunk_size).sum(axis=1)
+            upscaled_size = numpy.zeros_like(too_small_dests)
+            upscaled_size[::chunk_size] = chunks
```
zptro, line 100: This is the work of Copilot, with some revisions by me. Good job, Copilot!

zptro:
This does not have the sought effects for work tours, so we close this one.

PR #229, opened 2025-05-07
Hotfix/Add missing within zone costs to mode-destination choice

EliasMarttunen:


Scripts/parameters/demand/hb_edu_student.json
```diff
@@ -97,11 +97,20 @@
             },
             "log": {
                 "attraction_size": 1,
-                "cost": -0.28318
+                "cost": -0.28318,
+                "transform": -0.1
```
zptro, line 101: Because "cost" has been moved to "transform", it should not be here as well. We also need to apply the correct parameter.

```suggestion
                "transform": -0.28318
```

Scripts/parameters/demand/hb_leisure.json
```diff
@@ -97,11 +97,20 @@
...
                 "attraction_size": 1,
-                "cost": -0.59043
+                "cost": -0.59043,
+                "transform": -0.1
             },
             "attraction_size": {
                 "hospitality": 1,
                 "recreation": 1.49347
+            },
+                        "transform": {
```
zptro, line 107: Check the indentation.

Scripts/parameters/demand/hb_business.json
```diff
@@ -95,10 +95,19 @@
...
             "attraction_size": {
                 "workplaces": 1
+            },
+            "transform": {
+                "attraction": {
+                    "cost*within_zone": 1
+                },
+                "impedance": {
+                    "cost": 1
+                }
```
zptro, line 110: There is no log-cost parameter in the non-home-based models, so we should not add the transformation in them.

zptro (changes requested):

attesn:
@EliasMarttunen This Pull Request could have some written out name instead of short branch name, so git logs are easy to read in future. Could you change PR name to "Add missing within zone costs to mode-destination choice"?

attesn:
@zptro Would transforming impedance matrix diagonals in `purpose.transform_impedance` be more inefficient solution? I'd find it more understandable as impedance matrices entering choice models would be in same as in estimations. In this case it probably also requires less new code.

zptro:
> @zptro Would transforming impedance matrix diagonals in `purpose.transform_impedance` be more inefficient solution? I'd find it more understandable as impedance matrices entering choice models would be same as in estimations. In this case it probably also requires less new code.

Well, I actually removed more code than I added. The code handling the transformation is 6 rows.

I guess this is a matter of taste. I have had in mind trying to refactor `purpose.transform_impedance` a little bit, and if this fits nicely into that, we could also move it there later.

Scripts/tests/unit/test_logit.py
```diff
@@ -29,6 +29,8 @@ class Purpose:
         zi = numpy.array(INTERNAL_ZONES + EXTERNAL_ZONES)
         zd = ZoneData(ZONEDATA_PATH, zi, "uusimaa")
         zd["car_users"] = pandas.Series(0.5, zd.zone_numbers)
+        zd["cost"] = pandas.Series(0.0, index=zi)
+        zd["within_zone"] = pandas.Series(1, index=zi)
```
zptro, line 33: Good that you added "cost". But I think this should without adding "within_zone".

zptro (approved):

PR #231, opened 2025-05-14
Remove unnecessary transport_classes list and fix demand matrix validation

zptro:
Demand matrices are output to omx based on what is  in `assignment_modes`. To make it clearer and less error prone, now also base demand is read for `assignment_modes`. This makes `transport_classes` list unnecessary.

Also remove requirement for car and truck matrices from TransitAssignmentPeriod ("iltatunti").


EliasMarttunen (approved):

zptro:
Conflicts impossible to resolve according to Github. Started from clean branch in #254.

PR #234, opened 2025-05-16
Update long tour demand model parameters

attesn:
Generation and mode-destination choice parameters updated:
* New tour building method that matches total number of tours better to single day travel diary.
* Remove airplane from mode choice set where number of observations was less than ten (work, private day).


zptro (approved):
Technically, this looks good. The effect of removing airplane mode on distance distribution, and in the end kilometers traveled, worries me a bit, but if you think we can do without them, then ok.

PR #235, opened 2025-05-16
Improve handling purpose creation

Konnde:
Functionalised purpose creation so that it can be used for both domestic and foreign purpose handling. Added new freight util module for this and upcoming other tools.


Scripts/utils/freight_utils.py
```diff
@@ -0,0 +1,39 @@
...
+    in parameters path.
+
+    Parameters
+    ----------
+    parameters_path : Path
+        Path object to estimation model type folder containing model parameter
+        json files
+    purpose_args : list[Zonedata, ResultsData, dict]
+        arguments to init FreightPurpose class containing instance of Zonedata,
+        ResultsData and costs unit dictionary
```
zptro, line 20: I do not see much benefits in handling the arguments in this implicit way (as a list). I would prefer having them as separate variables, then you will not have to do the list modifications in the for-loop and the code will be easier to read.

zptro (changes requested):
Looks good. Just one small suggestion.

Scripts/utils/freight_utils.py
Konnde: It's true that endorsing explicit means here gets rid of the more obscure list handling. I'll update this accordingly

PR #236, opened 2025-05-16
Handle zone system related conflicts during storing

Konnde:
This branch is first solution to targeting zone size conflicts between different systems (emme/utility calculation/cost calculation etc) in freight demand calc when there are external zones in koko suomi network (foreign country centroids). For passenger demand similar problems are handled elsewhere.

Storing demand is functionalised here so that it can be used for both domestic and foreign purposes in addition to assessing whether demand matrix needs padding for emme compatibility.

### DIFF
--- a/Scripts/freight_lem.py
+++ b/Scripts/freight_lem.py
@@ -14,7 +14,7 @@ from assignment.emme_assignment import EmmeAssignmentModel
 from assignment.emme_bindings.emme_project import EmmeProject
 from datahandling.matrixdata import MatrixData
 
-from utils.freight_utils import create_purposes, store_demand
+from utils.freight_utils import create_purposes, StoreDemand
 from datahandling.traversaldata import transform_traversal_data
 from parameters.commodity import commodity_conversion
 
@@ -39,7 +39,9 @@ def main(args):
     purps_to_assign = list(filter(lambda purposes: purposes[0] in
                                   list(purposes), args.specify_commodity_names))
     ass_model.prepare_freight_network(costdata["car_cost"], purps_to_assign)
-    
+    store_demand = StoreDemand(ass_model.freight_network, resultmatrices, 
+                               zonedata.all_zone_numbers, zonedata.zone_numbers)
+
     impedance = ass_model.freight_network.assign()
     for mtx_type in impedance.keys():
         for ass_class, mtx in impedance[mtx_type].items():
@@ -52,15 +54,13 @@ def main(args):
     
     total_demand = {mode: numpy.zeros([zonedata.nr_zones, zonedata.nr_zones])
                     for mode in param.truck_classes}
-    set_mtx_args = (ass_model.freight_network, resultmatrices, 
-                    zonedata.all_zone_numbers, zonedata.zone_numbers)
     for purpose in purposes.values():
         log.info(f"Calculating demand for domestic purpose: {purpose.name}")
         demand = purpose.calc_traffic(impedance)
         for mode in demand:
-            save_demand = True if purpose.name in args.specify_commodity_names else False
-            store_demand(*set_mtx_args, mode, demand[mode], 
-                         save_demand, "freight_demand_tons", purpose.name)
+            omx_filename = ("freight_demand_tons" if purpose.name 
+                            in args.specify_commodity_names else "")
+            store_demand.store(mode, demand[mode], omx_filename, purpose.name)
         if purpose.name in args.specify_commodity_names:
             ass_model.freight_network.save_network_volumes(purpose.name)
         ass_model.freight_network.output_traversal_matrix(set(demand), resultdata.path)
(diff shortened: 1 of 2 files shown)


Scripts/utils/freight_utils.py
```diff
@@ -46,3 +49,64 @@ def create_purposes(parameters_path: Path, zonedata: FreightZoneData,
...
+    Emme network and whether demand should be saved on drive. 
+
+    Parameters
+    ----------
+    freight_network : FreightAssignmentPeriod
+        freight assignment period object
+    resultmatrices : MatrixData
+        handle for I/O matrix data handling
+    all_zones : numpy.ndarray
+        all zones in Emme network
```
Konnde, line 67: Quite many input parameters here. Zone numbers could be dropped if assesing demand dimensions was evaluated on freight_lem side. Additionally save_demand could be omitted if resultmatrices was used as optional checker. I don't think this however would improve readability as one would need to handle when to give resultmatrices on freight_lem side.

Scripts/freight_lem.py
```diff
@@ -40,32 +39,36 @@ def main(args):
     purps_to_assign = list(filter(lambda purposes: purposes[0] in
                                   list(purposes), args.specify_commodity_names))
     ass_model.prepare_freight_network(costdata["car_cost"], purps_to_assign)
+    
     impedance = ass_model.freight_network.assign()
+    for mtx_type in impedance.keys():
+        for ass_class, mtx in impedance[mtx_type].items():
+            impedance[mtx_type][ass_class] = mtx[:zonedata.nr_zones, :zonedata.nr_zones]
```
Konnde, line 46: Currently external impedances are not required in any process so I'll just drop them here. This however may change later when I get more information from team members.

Scripts/freight_lem.py
```diff
@@ -40,32 +39,36 @@ def main(args):
...
     del impedance["cost"]
     impedance = {mode: {mtx_type: impedance[mtx_type][mode] for mtx_type in impedance
                         if mode in impedance[mtx_type]}
                         for mode in ("truck", "freight_train", "ship")}
     
-    total_demand = {mode: numpy.zeros([len(zone_numbers), len(zone_numbers)])
+    total_demand = {mode: numpy.zeros([zonedata.nr_zones, zonedata.nr_zones])
                     for mode in param.truck_classes}
+    set_mtx_args = (ass_model.freight_network, resultmatrices, 
+                    zonedata.all_zone_numbers, zonedata.zone_numbers)
```
zptro, line 56: What you *could* do is to define a small `Class`, and set these in `__init__`. Then `store_demand` and `assess_demand_dimensions` could utilize them directly.

Scripts/utils/freight_utils.py
```diff
@@ -46,3 +49,64 @@ def create_purposes(parameters_path: Path, zonedata: FreightZoneData,
             log.warn(f"Aggregated commodity class '{commodity_conversion[commodity]}' "
                       f"for commodity {commodity} not found in costs json.")
     return purposes
+
+def store_demand(freight_network: FreightAssignmentPeriod, resultmatrices: MatrixData, 
+                 all_zones: numpy.ndarray, zones: numpy.ndarray, 
+                 mode: str, demand: numpy.ndarray, 
+                 save_demand: bool, omx_filename: str, key_prefix: str = ""):
```
zptro, line 56: You could consider having `omx_filename` as optional, then the extra `save_demand` flag would not be necessary.

Scripts/utils/freight_utils.py
```diff
@@ -46,3 +49,64 @@ def create_purposes(parameters_path: Path, zonedata: FreightZoneData,
...
+    Returns
+    -------
+    numpy.ndarray
+        demand with/without zone padding
+    """
+    fill_mtx = demand
+    if demand.size != nr_all_zones**2:
+        fill_mtx = numpy.zeros([nr_all_zones, nr_all_zones], dtype=numpy.float32)
+        fill_mtx[:nr_zones, :nr_zones] = demand
+    return fill_mtx
```
zptro, line 112: Could you please add an empty row in the end of the file?

zptro (dismissed):
Looks great! Only some small remarks.

Scripts/utils/freight_utils.py
Konnde: I had same thoughts, so with the feedback from here I'll update accordingly

Scripts/freight_lem.py
Konnde: Good idea. I was somewhat purposefully avoiding OOP here, as I thought it would mix up class methods and functions in freight utils module. But in this suggestion I think it would be just fine as it would wrap the default arguments and the two methods nicely

zptro (approved):
Looks good!

PR #237, opened 2025-05-22
Systematically check all shares and calculate totals

zptro:
Replace the hard-coded population/workplace/household segment calculation with a general calculation of `"sh_type_x" * "total"`. Systematically check that no shares sum to more than one.

### DIFF
--- a/Scripts/datahandling/zonedata.py
+++ b/Scripts/datahandling/zonedata.py
@@ -284,7 +284,7 @@ def read_zonedata(path: Path,
         (Path(__file__).parent / "zone_variables.json").read_text("utf-8")
     )[data_type]
     aggs = {}
-    shares: Dict[str, Dict[str, List]] = {}
+    shares: Dict[str, Dict[str, List[str]]] = {}
     for func, cols in zone_variables.items():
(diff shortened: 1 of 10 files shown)


Scripts/datahandling/zonedata.py
```diff
@@ -64,26 +65,10 @@ def _add_transformations(self, data):
...
-                share_7_99 += pop_share
-            if col.startswith("sh_income"):
-                self[col.replace("sh_", "")] = data[col] * pop
-            if col.startswith("sh_wrk_"):
-                self[col.replace("sh_wrk_", "")] = data[col] * wp
-            if col.startswith("sh_hh"):
-                self[col.replace("sh_", "")] = data[col] * hh
-        self.share["share_age_7-99"] = share_7_99
+            if col.startswith("age"):
+                self[col + "_female"] = 0.5 * data[col]
```
attesn, line 70: If I'm correct these variables are not used anymore. They were part of long distance generation, but removed from latest versions.

attesn (approved):

attesn (approved):

PR #239, opened 2025-05-22
Add ton-eur information to purpose result summary

Konnde:
After successful external model piloting, a request to add product of cost matrix and ton demand matrix to freight model as a fixed result information was made. This PR is meant to address that request.

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -581,15 +581,30 @@ class FreightPurpose(Purpose):
         dict
             Mode (truck/train/...) : calculated demand (numpy 2d matrix)
         """
-        costs = {mode: {"cost": calc_cost(mode, self.costdata, impedance[mode])}
-            for mode in self.modes}
-        return_costs = copy(costs)
+        costs = self.get_costs(impedance)
         self.dist = costs["truck"]["cost"]
         nr_zones = self.zone_data.nr_zones
         probs = self.model.calc_prob(costs)
         generation = numpy.tile(self.zone_data[f"gen_{self.name}"], (nr_zones, 1))
         demand = {mode: (probs.pop(mode) * generation).T for mode in self.modes}
-        return demand, return_costs
+        return demand
+
+    def get_costs(self, impedance: dict):
+        """Fetches calculated costs for each mode in model's mode choice.
+
+        Parameters
+        ----------
+        impedance : dict 
+            Mode (truck/train/...) : dict
+                Type (time/cost/dist) : numpy 2d matrix
+
+        Returns
+        -------
+        dict
+            Mode (truck/freight_train/...) : cost : numpy.ndarray
+        """
+        return {mode: {"cost": calc_cost(mode, self.costdata, impedance[mode])}
+                for mode in self.modes}
 
     def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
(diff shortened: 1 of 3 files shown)


Scripts/datatypes/purpose.py
```diff
@@ -583,12 +583,13 @@ def calc_traffic(self, impedance: dict):
         """
         costs = {mode: {"cost": calc_cost(mode, self.costdata, impedance[mode])}
             for mode in self.modes}
+        return_costs = copy(costs)
         self.dist = costs["truck"]["cost"]
         nr_zones = self.zone_data.nr_zones
         probs = self.model.calc_prob(costs)
```
Konnde, line 589: contents of costs variable are popped at different phases of logit calculation depending whether given model type was mode>dest or dest>mode, so we are unable to return that cost information and thus need to make a copy. This is slightly problematic as needlessly fills memory during koko suomi (over 9400 centroids) model run.

Scripts/datatypes/purpose.py
```diff
@@ -581,15 +581,31 @@ def calc_traffic(self, impedance: dict):
...
+        costs = self.get_costs(impedance)
         self.dist = costs["truck"]["cost"]
         nr_zones = self.zone_data.nr_zones
         probs = self.model.calc_prob(costs)
         generation = numpy.tile(self.zone_data[f"gen_{self.name}"], (nr_zones, 1))
         demand = {mode: (probs.pop(mode) * generation).T for mode in self.modes}
         return demand
 
+    def get_costs(self, impedance: dict):
+        """Fetches calculated costs for each mode in model's mode choice.
```
Konnde, line 593: Added getter so that we don't have to carry copy of costs from calc_traffic back to freight_lem in order to use them in writing purpose summary

Scripts/freight_lem.py
```diff
@@ -83,24 +84,31 @@ def main(args):
...
     mode_tons = [numpy.sum(demand[mode])+0.01 for mode in modes]
     shares_tons = [tons / sum(mode_tons) for tons in mode_tons]
     mode_ton_dist = [numpy.sum(demand[mode]*impedance[mode]["dist"])+0.01 for mode in modes]
     shares_mileage = [share / sum(mode_ton_dist) for share in mode_ton_dist]
+    costs = purpose.get_costs(impedance)
+    ton_costs = []
+    for mode in modes:
+        costs[mode] = numpy.where(costs[mode]["cost"] == numpy.inf, 
+                                  0, costs[mode]["cost"])
+        ton_costs.append(numpy.sum(costs.pop(mode)*demand[mode]))
```
Konnde, line 103: For freight transit modes, the impact of inf cost impedances for some OD pairs in demand calculation causes the demand to be 0 so here we can just replace infs with zeros.

zptro, line 103:
```suggestion
    costs = {mode: c["cost"] for mode, c in purpose.get_costs(impedance).items()}
    for cost in costs.values():
        cost[cost == numpy.inf] = 0
    ton_costs = [numpy.sum(costs.pop(mode)*demand[mode]) for mode in modes]
```

zptro (approved):
Looks good! Here is just one small code style suggestion.

Scripts/freight_lem.py
Konnde: This is actually nice, great suggestion

zptro (approved):

PR #241, opened 2025-05-23
..-separated intervals instead of hyphen

zptro:
This is easier to copy-paste into excel without date interpretation.

### DIFF
--- a/Scripts/datatypes/histogram.py
+++ b/Scripts/datatypes/histogram.py
@@ -8,8 +8,8 @@ class TourLengthHistogram:
     _u = numpy.array(intervals[1:])
 
     def __init__(self, name):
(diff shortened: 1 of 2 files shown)


attesn:
Downside is that this will break compatibility to well-used plotting scripts in other repos.

zptro:
@attesn what about this more modest change?

attesn:
Is this causing a lot of extra work? Even modest change is surprisingly tricky, because we'll need to adjust multiple repositories or add quickfixes to many plotting scripts. One of those cases, where our data production should be better optimized.

zptro:
@attesn for everyone using excel for analyzing results (i.e., a large share of the future users of the model system), the current format means manual work needs to be done _everytime_ the histogram file is opened.

If not for this problem, excel with Pivot Table can be used to produce nice graphs in a few minutes.

attesn (approved):

PR #242, opened 2025-05-26
Weight congested car time (time - free_flow_time) by 1.5

zptro:
In this solution, route choices are not affected by the extra weight on congested time (as that would require us to change the volume-delay functions, affecting actual link speeds). However, the time variable, which is fed to the demand model, is directly modified. This means that the current solution also affects cost-benefit analyses. This is perfectly in line with [ASEK 8.0](https://bransch.trafikverket.se/contentassets/0e5777a6301e4134a6e8365fc20c0e0e/asek-8.0-2-april-2024.pdf), but not with the current version of [Hankearvioinnin yksikköarvot](https://aineistot.vayla.fi/ava/Julkaisut/Vaylavirasto/vo_2024-44v2_hankearvioinnin_yksikkoarvot_2022_web.pdf). Maybe we should try to make it part of the next version?

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -354,6 +354,8 @@ class AssignmentPeriod(Period):
                         link.volume_delay_func = roadclass.volume_delay_func
                 link.data1 = roadclass.lane_capacity
                 link.data2 = roadclass.free_flow_speed
+                link[param.free_flow_time_attr] = (60 * link.length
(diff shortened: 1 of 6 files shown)


attesn (approved):

attesn:
car_tours_subarea
Results with new impedance matrices. Significant changes only in Helsinki downtown area, which probably is because there is not much congestion elsewhere.

PR #246, opened 2025-06-04
Require minimum number of stored car times in stored-speed assignment

zptro:
Less than 50,000 missing car-time values is still OK, but issues a warning with list of all missing values. If more than 50,000 car-time values are missing, a whole sub-model is probably missing, so then we throw an error.

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -349,7 +349,6 @@ class AssignmentPeriod(Period):
         }
         park_and_ride_mode = network.mode(param.park_and_ride_mode)
         car_time_zero = []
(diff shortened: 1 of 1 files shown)


attesn (approved):
Seems good. Threshold could be lower on my opinion (1000-2000), as whole Helsinki downtown has 10000 links and city of Lahti only 3500 links.

zptro:
> Seems good. Threshold could be lower on my opinion (1000-2000), as whole Helsinki downtown has 10000 links and city of Lahti only 3500 links.

We could lower the threshold when we have made sure transferring travel times based on node numbers is more stable than it has been previously.

PR #247, opened 2025-06-06
Read in freight marine transit line attributes

Konnde:
Tool implementation for fetching network attributes for freight marine modes (ship modes between Finnish border points and foreign cluster border points). Read attributes are converted to ship mode specific impedance matrices. Additionally, information about origin and destination dimensions is constructed to be used further in cost calculation.


Scripts/freight_lem.py
```diff
@@ -80,6 +80,8 @@ def main(args):
                                resultdata, costdata["freight"])
     for purpose in purposes.values():
         log.info(f"Calculating demand for foreign purpose: {purpose.name}")
+        ship_impedances = ReadShipImpedances(zonedata.all_zone_numbers, "export")
+        impedance["ship"], ext_zones = ship_impedances.form_matrices(ass_model.freight_network)
```
Konnde, line 84: For now impedances are set here as a temporary solution. Next development steps include parsing trade omx from which we are able to fetch import and export specific commodities and update this block accordingly

PR #250, opened 2025-06-06
Read ship impedances (FreightAssignmentPeriod implementation)

zptro:
This is an alternative implementation of #247. I think it makes sense to integrate the methods in `FreightAssignmentPeriod`, on which they operate. I also simplified some of the lookups.


Scripts/freight_lem.py
```diff
@@ -80,7 +80,9 @@ def main(args):
                                resultdata, costdata["freight"])
     for purpose in purposes.values():
         log.info(f"Calculating demand for foreign purpose: {purpose.name}")
-    
+        imp, origs, dests = ass_model.freight_network.read_ship_impedances(
+            is_export=True)
+        impedance["ship"] = imp
```
Konnde, line 85: imp attributes now appear as ut1 and ut2. I would say that that we add mapping to either in assignment.py or read_ship_impedances framework for ut1 to be "dist" and ut2 to be "frequency" so that we can reference imp dist directly in calc_ship_costs and frequency later in route choice.

Konnde (commented):
Very nice refactoring! Makes the tool more concise and better utilizes existing assignment period instead supporting freight utils module. Tested this with expanded test network and found no problems. Just one comment for future update needs

Scripts/freight_lem.py
zptro: Good point! I fixed this now.

PR #252, opened 2025-06-11
Expand freight cost calculations routines with foreign trade

Konnde:
This branch is clean version of feat/update_cost_calc which has been used in drafting this structure.

Main feature updates include:
1. Update cost calculation within Finland (domestic) to match calculation restructuring from external consultant analysis project
2. Introduce foreign cost calculation routines. For truck and train we can utilize already existing functions. For marine ships, new cost minimization routine is added


Scripts/utils/freight_costs.py
```diff
@@ -155,5 +195,113 @@ def get_aux_cost(unit_costs: Dict[str, Dict],
...
+    mask = numpy.ones((len(ext_origin.values()), len(ext_dest.values())))
+    if set(ext_origin.keys()).issubset(set(port_draught.keys())):
+        origin_based = True
+        fin_ports = list(ext_origin)
+    else:
+        origin_based = False
+        fin_ports = list(ext_dest)
+    for index, port in enumerate(fin_ports):
+        if draught > port_draught[port]:
+            mask[index][0:] = numpy.inf
```
zptro, line 285:
```suggestion
    for index, port in enumerate(fin_ports):
        if draught > port_draught[port]:
            mask[index, :] = numpy.inf
```

Scripts/utils/freight_costs.py
```diff
@@ -155,5 +195,113 @@ def get_aux_cost(unit_costs: Dict[str, Dict],
...
+    none_mtx = numpy.full((len(origs), len(dests)), None, dtype="float64")
+    ship_info = {}
+    for mode in unit_costs["ship"].keys():
+        if mode == "domestic_vessel":
+            continue
+        ship_info[mode] = {
+            "cost": none_mtx.copy(),
+            "draught": none_mtx.astype("object")
+        }
+        for ship_draught, ship_params in unit_costs["ship"][mode].items():
```
zptro, line 239: I do not understand what you are iterating over here. In example file `costdata.json` we have the parameters ("time", "dist") directly under "domestic_vessel". Could you add some example data for foreign freight to `costdata.json`?

Scripts/utils/freight_costs.py
```diff
@@ -155,5 +195,113 @@ def get_aux_cost(unit_costs: Dict[str, Dict],
...
+            draught = int(ship_draught[:-1])
```
zptro, line 240: We could just have it as an integer in unit costs?

Scripts/utils/freight_costs.py
```diff
@@ -85,50 +101,72 @@ def calc_rail_cost(unit_costs: Dict[str, Dict],
...
+    """Fetch domestic freight ship based costs. 
 
     Parameters
     ----------
     unit_costs : Dict[str, Dict]
         Freight mode (truck/freight_train/ship) : mode
             Mode (other_dry_cargo...) : draught
-                draught (4m/7m/9m) : unit cost name
+                draught (4m...) : unit cost name
                     unit cost name : unit cost value
```
zptro, line 122: This seems outdated

Scripts/utils/freight_costs.py
```diff
@@ -155,5 +195,113 @@ def get_aux_cost(unit_costs: Dict[str, Dict],
...
+        Destination border id (FIHEL/SESTO...) : str
+            Centroid id : int
+
+    Returns
+    -------
+    dict
+        Marine ship type (general_cargo/container ship...) : matrix
+            Mtx type (cost/mode/draught) : numpy.ndarray
+    """
+    none_mtx = numpy.full((len(origs), len(dests)), None, dtype="float64")
```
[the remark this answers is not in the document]
Konnde: I have understood that while you can have nan values in an array, there is no array dtype "None". Additionally, the aim is to insert costs type float into the matrix anyway. But like I said, this base matrix could now be inf mtx and inf is classified as dtype float.

Scripts/utils/freight_costs.py
Konnde: I think you're correct, and I have made splicing error here. Yes, the assumption is that the origin and destination matrices do not have equal size.

Scripts/utils/freight_costs.py
Konnde: It seems that I had forgotten to push updated test costdata with this marine ship layer. Now fixed.

Scripts/utils/freight_costs.py
Konnde: Yes we can. The draught key has been string for quite long time now, and this development step would be a good place to change the dtype from string to int as we are now dealing with several mappings other than unit costs that require draught info.

Scripts/utils/freight_costs.py
Konnde: Will update docstring accordingly along draught dtype conversion updates.

Scripts/utils/freight_costs.py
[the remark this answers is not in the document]
Konnde: Yeah you're correct. While the dimensions in these border-border matrices are fraction in comparison to whole Finland, there is no real need for these too to be 64bit. Will update

zptro (approved):
Yes, now this can be merged!

PR #253, opened 2025-06-12
New calibration after stored speeds were fixed

EliasMarttunen:


zptro (changes requested):
This looks good, but I changed the base branch, as we insert the calibration parameters only in main branch. Could you resolve the conflicts in the parameter files?

PR #256, opened 2025-06-17
Add script that creates empty EMME project

zptro:
This is useful if the user already has an existing emmebank, and would like to link it to the project.


zptro:
> > This is useful if the user already has an existing emmebank, and would like to link it to the project.
> 
> Trying to figure out how this is initiated; does the emmebank file have to be located within the 'Projektin nimi" directory or will there be new path option where emmebank location can be given?

So, when you have your new project (which is just an .emp-file and some settings files related to it), you can open it with EMME and link any existing emmebank file to it. It does not matter what the location of the file is, the .emp-file has the path to it.

Konnde (approved):

PR #259, opened 2025-06-19
Overnight segmentation in demand models

attesn:
Use overnight segmentation instead of distance boundaries. Estimation of daily tours are not updated here, although boundary is increased. That will be done in next PR as we updated all demand models with new impedance matrices.


attesn:
@zptro I cannot see why tests are failing to matrix import, as nothing related to that was not changed.

zptro:
> @zptro I cannot see why tests are failing to matrix import, as nothing related to that was not changed here. Everything was working ok before merging `dev`.

It is the beeline matrix directly from the assignment model that is not compatible with demand model bounds. When you add it to zonedata, you need to slice it according to `ZoneData.zone_numbers`.

Konnde (approved):
I'm not very familiar with the passenger/travel changes so I mostly reviewed styling and general outlook.

PR #266, opened 2025-08-06
Fix EMME freight assignment test broken in #260

zptro:


Konnde (approved):
Good catch. Seems that I missed this when I updated transform_traversal_data to return aux demand instead of aggregated truck + aux demand

PR #269, opened 2025-08-07
Add Helsinki - Jyvaskyla LOS logging

zptro:
Log travel time, cost, dist, etc for all modes of transport between Helsinki and Jyväskylä. Also skip zero travel time ODs (i.e., matrix diagonals) in OD speed descriptive logging.

### DIFF
--- a/Scripts/utils/validate_assignment.py
+++ b/Scripts/utils/validate_assignment.py
@@ -16,10 +16,26 @@ def divide_matrices(mtx1: numpy.ndarray,
     v = [round(numpy.quantile(mtx, q)) for q in [0.00, 0.50, 1.00]]
     log.debug(f"{description} (min, median, max) {v[0]} - {v[1]} - {v[2]}")
 
-def output_od_los(mtx: numpy.ndarray,
+def output_od_los(los_mtx: numpy.ndarray,
                   mapping: Dict[int, int],
                   mtx_type: str,
                   mode: str):
-    los = int(mtx[mapping[202], mapping[17278]])
+    """Log LOS between Helsinki and Jyvaskyla.
+
+    Parameters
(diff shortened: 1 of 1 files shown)


Scripts/utils/validate_assignment.py
```diff
@@ -0,0 +1,41 @@
...
+            Zone number
+        value : int
+            Matrix index
+    mtx_type : str
+        Type (time/cost/dist/...)
+    mode : str
+        Assignment class (car_work/transit_leisure/...)
+    """
+    los = int(los_mtx[mapping[202], mapping[17278]])
+    los = f"{los//60}h {los%60}min" if "time" in mtx_type else los
```
attesn, line 40: Should we have "km" and "eur" units as well for readability?

Scripts/utils/validate_assignment.py
```diff
@@ -0,0 +1,41 @@
...
+        key : int
+            Zone number
+        value : int
+            Matrix index
+    mtx_type : str
+        Type (time/cost/dist/...)
+    mode : str
+        Assignment class (car_work/transit_leisure/...)
+    """
+    los = int(los_mtx[mapping[202], mapping[17278]])
```
attesn, line 39: Does this work if we're using submodels?

attesn (approved):
This is good. 

Some further ideas:
* Could we use some zone pair where the effect of (missing) congestion could be seen?
* Could we have 2016 times stored and warn if suspiciously different from that?

Scripts/utils/validate_assignment.py
zptro: These zones are set as their region's centers, so they remain in all aggregations.

PR #271, opened 2025-08-19
Add possibility for extra area dummy variables

zptro:
The dummy variable is an area (i.e., a set of zones sharing a common aggregate area name) in one of the "aggregate_results_*" columns in zone data, and its parameter is defined in json. The parameter affects mode choice for persons living in the affected zones.


attesn (dismissed):
Nice! Probably the best review is done together with case study, but PR can be already merged.

PR #274, opened 2025-08-20
Minor fixes to demand model parameters

attesn:
Alogit estimates export fixed:
* Estimates were not from last iteration round (log file), but previous one (f12 file).
* Multipart size factors were not corrected correctly. This does affect to distance distribution of free flow model.

### DIFF
--- a/Scripts/parameters/demand/hb_business.json
+++ b/Scripts/parameters/demand/hb_business.json
@@ -38,23 +38,23 @@
     },
     "distance_boundaries": {
         "car_work": [
-            0,
+            0.1,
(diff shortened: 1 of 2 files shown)


Scripts/parameters/demand/hb_business.json
```diff
@@ -38,23 +38,23 @@
     },
     "distance_boundaries": {
         "car_work": [
-            0,
+            0.1,
```
zptro, line 41: What is the meaning of this?

zptro (approved):

PR #278, opened 2025-08-25
Move walk and bike assignment to `OffPeakPeriod`

zptro:
Motivation:
`AssignmentPeriod` is typically created as two objects: "aht" and "iht". Until now, they have both had full assignment of walk and bike, but only "iht" LOS has been used in demand calculation (because "aht" and "iht" LOS are identical). It is clearer and more memory efficient to have it in `OffPeakPeriod`, which has only one object: "pt".

Implications:
- For running the model normally (without `DO_NOT_USE_EMME`), nothing should change.
- If running with `DO_NOT_USE_EMME`, walk distance needs to be moved from `dist_iht.omx` to `dist_pt.omx`. @attesn 
- After new LOS matrices have been produced, model estimation needs to be changed so that it picks walk and bike distance from "pt" instead of "iht". @attesn


attesn:
@EliasMarttunen @zptro Is there a plan at which point matrices are updated to Sharefile?

PR #279, opened 2025-08-25
Use continuous size factor for overnight tours

attesn:
Specification produces similar results, but without steps and less variables.

### DIFF
--- a/Scripts/travel_iteration.py
+++ b/Scripts/travel_iteration.py
@@ -281,10 +281,7 @@ class ModelSystem:
         mtx = self.ass_model.beeline_dist
         idx = numpy.where(numpy.isin(self.zdata_forecast.zone_numbers, self.zone_numbers))[0]
         mtx = mtx[idx[:, None], idx]
-        self.zdata_forecast["beeline_10km"] = mtx<10
-        self.zdata_forecast["beeline_100km"] = (mtx>10) & (mtx<100)
-        self.zdata_forecast["beeline_200km"] = (mtx>100) & (mtx<200)
-        self.zdata_forecast["beeline_9999km"] = mtx>200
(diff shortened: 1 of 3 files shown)


zptro:
What bothers me with this is that we are no longer estimating the impact of beeline distance on destination choice, but fixing it according to an arbitrarily chosen function. Does not the sharp increase of the step-wise parameters imply that we should use a more sharply increasing function (e.g., non-logarithmic beeline distance)?

Scripts/travel_iteration.py
```diff
@@ -281,10 +281,7 @@ def assign_base_demand(self,
         mtx = self.ass_model.beeline_dist
         idx = numpy.where(numpy.isin(self.zdata_forecast.zone_numbers, self.zone_numbers))[0]
         mtx = mtx[idx[:, None], idx]
```
zptro, line 283: This is not working as intended, because the  order of `self.zdata_forecast.zone_numbers` and `self.zone_numbers` is wrong. What it does now is actually equal to the much simpler:
```
idx = len(self.zdata_forecast.zone_numbers)
mtx = mtx[:idx, :idx]
```

I think that what you wanted to achieve is best expressed like this, because it is unnecessary to transform the boolean index into an integer index:

```
idx = numpy.isin(self.zone_numbers, self.zdata_forecast.zone_numbers)
mtx = mtx[numpy.ix_(idx, idx)]
```

zptro (changes requested):
The actual PR is fine, as long as we keep in mind that it could be worthwhile testing `beeline_log` as a separate variable.

I just have a suggestion related to the previous PR that was merged when I was on vacation.

Scripts/travel_iteration.py
zptro: We could do it like that.

zptro (approved):

PR #284, opened 2025-09-09
Update time period transformations for demand

jmutikainen:
- Updated values for impedance share
- Updated values and format for volume factors (1/x => x)
- Updated values and format for external demand (one value for time period and direction => one value for time period)


attesn (approved):
Two step process looks good for me.

Impedance share labels causes some confusion for me. I think they actually refer to "ah" not "aht".

PR #285, opened 2025-09-10
Update parameters for mode-destination choice

attesn:
- New congested LOS matrices with corrected base demand.
- Impedance shares updated.
- Closed tours with one trip fixed.
- Area based transit dummy variables removed.


attesn:
Some validation results:
* Mode shares are similar to previous version in national scale, but subarea validation are now logical. Car driver is overestimated in all urban areas. HB Work has car driver share underestimated, but this was problem already before.
* Values-of-time increases compared to previous estimation. Home-based Other Shop has false value-of-times as linear cost component becomes very close to zero and statistically insignificant.

PR #296, opened 2025-10-03
Update create_emmebank.py

juliuskroetzl:
Increase number of centroids for alueelliset_osamallit from 3400 to 3800


zptro (commented):
3400 -> 4100? How can the number of zones have changed that much?

zptro:
> Yes, itä-suomi submodel has now 4051 zones.

Why?

PR #301, opened 2025-10-17
Use  logit generation model in place of of tour combinations

attesn:
Tour generation models are currently independent logit models for each purpose and it is uncertain if tour combination model is to be used. Therefore replacing applied use of tour combination model with generation logit model. 

* Tour generation logit is very similar to car ownership, so using same logit model class.
* Generation model selection requires some changes. Here I'm a bit unsure what would be best option.


zptro (approved):
This is a really nice change, which simplifies further development of the models!

PR #302, opened 2025-10-21
Set long-dist mode unavailable if main mode less than half of dist

zptro:
Long-distance transit (i.e., train, coach, airplane) _assignment_ is performed as usual, enforcing usage of the main mode, but without restriction of the extent of usage. Hence, the distance traveled by the main mode can in some cases be a very short part of the total distance traveled. In these cases, we now set the long-distance transit mode as unavailable in _demand calculation_ by setting its travel time to 999999.

Also use the same analysis method for output of transit_work and transit_leisure demand using j and r modes (train) in EMME.

TODO:
- Remove dist matrix before demand calculation

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -217,6 +217,7 @@ class AssignmentPeriod(Period):
         self._check_congestion()
         del mtxs["dist"]
         del mtxs["toll_cost"]
+        del mtxs["train_users"]
         return mtxs
 
     def end_assign(self) -> Dict[str, Dict[str, numpy.ndarray]]:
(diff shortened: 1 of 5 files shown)


EliasMarttunen (dismissed):

attesn:
Does this apply to airplane matrices? I'd guess that for most of Lapland the access time is larger than airplane in-vehicle time, as airplane is so much faster than transit or car. Would it be possible to use in-vehicle distance or exclude airplanes from this?

attesn:
@EliasMarttunen Can you run model again and update matrices to Sharefile? Matrices are a bit different after comparison was changed to distance.

attesn (commented):
I think we can close this PR as #308 is better development option.

PR #306, opened 2025-10-28
License-dependent car-ownership model implementation

zptro:
- Add new zone input variable `sh_adult_license` representing the share of adult population having a driver's license
- Calculate artificial shares of households with given number of adults, children and licenses
- Change car-ownership model to be dependent on the number of adults and licenses in household


zptro:
@attesn What are the calibration targets for the car-ownership model now when we have it for license-owners-only? @ellimaijavirtanen could proceed to calibration as soon as we have them.

attesn:
> @attesn What are the calibration targets for the car-ownership model now when we have it for license-owners-only? 

You can get Tilastokeskus stats by running `cars_stats.R` in prepare-zone-data repository.

zptro:
> > @attesn What are the calibration targets for the car-ownership model now when we have it for license-owners-only?
> 
> You can get Tilastokeskus stats by running `cars_stats.R` in prepare-zone-data repository.

I see! To make it clear, the new model outputs car ownership shares for the following groups:
- 1 adult, 1 license households
- 2 adults, 1 license households
- 2 adults, 2 licenses households

We do not have any Tilastokeskus car-ownership stats for these groups, so I assumed without deeper thinking that we would calibrate the ASCs against HLT data. Do you think we should
1. calibrate against HLT or
2. wait for new TK data?

If we choose to wait for the new TK data, I will mark this PR as ready for review, as I think this branch should be merged ASAP now that the new functionality is implemented. We can then calibrate it later.

attesn (changes requested):
There was bias in municipality level validation of car-ownership. We'll need to update zonedata and calibrate model again after basedata is fixed.

attesn:
Zonedata bias was fixed, but validation shows that households without car are still systematically underestimated. I've used 2016 zonedata and 2018 LOS-matrices from Valma Sharefile.

attesn (approved):
Validation with Tilastokeskus data shows some overestimation of one car ownership and underestimation of two cars ownership. However, differences are most likely caused by differences in validation data sources.

PR #307, opened 2025-10-31
Loop over tour purposes only once in demand calculation

zptro:
Now, when tour generation is no longer shared between tour purposes, we can perform the whole calculation [logsum -> tour generation -> mode and destination choice] in one shot for each tour purpose, and thereby we no longer need to store large probability matrices in `TourPurpose`.

I also deactivated the separate handling of soft modes. This can be reactivated later if we decide to start using a denser zoning  for them.


attesn (commented):
Very good update!

PR #309, opened 2025-11-03
Update main including calibration of mode choice constants

EliasMarttunen:


attesn:
We could change title to "Calibration of mode choice constants". No need to keep date in title, as all commits have time stamp automatically.

attesn (dismissed):
Calibration constants are reasonable when compared to observed and modelled (without calibration) mode shares. 
* WB Business and OB Other required large contants. Here the primary tour mode choice is most powerful explanatory factor  and therefore errors are partly caused by errors in other purposes. However, these have minor role in total demand.
* HB Work worries me most as walk and bike need to be adjusted down. There is something not working correctly with work tours. As problem is walk and bike together, this might caused by within zone impedances.

attesn (approved):

PR #316, opened 2025-11-07
Fix bugs in long-distance assignment

zptro:
- Remove vans from long-distance assignment
- Fix CAR_END_ASSIGNMENT_ONLY for long-distance assignment


jalovaj2:
Otherwise looks good!

PR #318, opened 2025-11-11
Change convergence criteria logic

zptro:
Change in accordance to HSL findings.


samakinen (approved):
This is a good change. Applying the same modification to the Helmet model system resolved the issue of iterations ending prematurely.

PR #320, opened 2025-11-19
Prevent duplicate tour purposes in travel iteration

attesn:


zptro (approved):
It would be quite simple to include the duplicate purpose name in the error message , but the user would have to search trough the json files anyway to find the duplicate.

PR #308, opened 2025-10-31
Draft long-distance transit assignment

zptro:
- Remove `l` mode (airplane) from `transit_work` and `transit_leisure` assignment
- Enable `transit_work` and `transit_leisure` assignment in `WholeDayPeriod`
  - No specific mode is enforced
- Remove separate `train` and `coach` assignment
- Introduce new joint `train` and `coach` car-access assignment
  - Either `j` or `e` mode enforced (no enforcement? `j` enforced?)

PR #325, opened 2025-12-09
Output train_users matrix

zptro:
Output of transit_work and transit_leisure demand using j and r modes (train) in EMME. Subset of #302

### DIFF
--- a/Scripts/assignment/datatypes/transit.py
+++ b/Scripts/assignment/datatypes/transit.py
@@ -153,7 +153,7 @@ class TransitMode(AssignmentMode):
                 "avg_boardings": self.num_board.id,
             },
         }]
-        # For daily tours, use main_inv_time for analyzing train usage
+        # For daily tours, use main_mode_dist for analyzing train usage
         self.transit_result_specs.append({
             "type": "EXTENDED_TRANSIT_MATRIX_RESULTS",
             subset: {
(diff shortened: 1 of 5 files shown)


Scripts/assignment/datatypes/transit.py
```diff
@@ -152,6 +153,14 @@ def _add_matrix_specs(self, modes):
                 "avg_boardings": self.num_board.id,
             },
         }]
+        # For daily tours, use main_inv_time for analyzing train usage
```
jalovaj2: is `main_inv_time` in this comment the same thing as `main_mode_dist` that is actually used in `transit_result_specs`?

zptro: Good point, this comment is not up-to-date, I will update it

jalovaj2:
We noticed with @attesn  that some version of this branch (I'm not exactly sure which) produced train_users_vrk.omx -file with no matrices inside, only the lookup. I couldn't instantly find a bug in this Pull Request but may be worth checking it out with @attesn whether the file was created by a previous version of this PR or whether the bug still exists.

zptro:
> We noticed with @attesn that some version of this branch (I'm not exactly sure which) produced train_users_vrk.omx -file with no matrices inside, only the lookup. I couldn't instantly find a bug in this Pull Request but may be worth checking it out with @attesn whether the file was created by a previous version of this PR or whether the bug still exists.

I think what you refer to is what the long-distance model currently outputs. Currently the train users in the long-distance model is a no-brainer: they are in the "train" matrix in demand_vrk.omx. The train_users-matrices on the other hand are produced only for transit_work and transit_leisure in the short-distance model run.

Good that you mentioned this, because in the future (#308) we will also have train users in the "airplane" matrices. So we should add output for them as well.

jalovaj2 (approved):
Looks good!

PR #328, opened 2025-12-18
Update fetching impedances for marine ship modes

Konnde:
Fix bad implementation where marine transit lines were fetched and processed from emme api multiple times during reading impedance information. Now they're read and processed once.

### DIFF
--- a/Scripts/assignment/freight_assignment.py
+++ b/Scripts/assignment/freight_assignment.py
@@ -202,13 +202,12 @@ class FreightAssignmentPeriod(AssignmentPeriod):
         return port_ids
 
     def _get_marine_transit_lines(self, is_export: bool) -> dict:
-        """Get list of marine transit lines depending whether data fetching
-        is done for exporting or importing.
+        """Fetch list of marine transit lines.
 
         Parameters
         ----------
         is_export : bool
-            Whether data should be fetched for export (True) or import (False)
+            Fetch export specific transit lines. Else import specific
         
         Returns
         -------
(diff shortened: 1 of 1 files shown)


Scripts/assignment/freight_assignment.py
```diff
@@ -198,22 +201,50 @@ def _filter_border_points(self, border_data: dict) -> Dict[str, int]:
...
+                continue
+            
+            id_split = line.id.split("_")[0]
+            try:
+                int(id_split)
+                continue
+            except ValueError:
+                starts_in_fi = line.id[0:2] == "FI"
+                if ((starts_in_fi and is_export) or (not starts_in_fi and not is_export)):
+                    mode_transit_lines[line.mode.id].append(line)
```
zptro: This logic seems a bit backwards. The thing inside the exception is what we normally want to perform, and the value being an `int` is what would break it, right? Could this be expressed more straightforward, for instance using `if line.id[0:2].isalpha()` or similar?

zptro (changes requested):
Looks really good! Just one minor suggestion:

Scripts/assignment/freight_assignment.py
Konnde: Ah, good idea. I will make it this more elegant.

Partly the reason we even need to check this is because network builder currently creates those link specific transit lines similar to W-mode but for the marine modes. I have informed about this and it most likely gets fixed later.

zptro (approved):

PR #329, opened 2025-12-19
Improve freight impedance handling

Konnde:
Run foreign route choice model first, and move responsibilities from valma freight to other functions and methods. Enables better memory management.

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -704,8 +704,7 @@ class FreightPurpose(Purpose):
         demand = {mode: (probs.pop(mode) * generation).T for mode in self.modes}
         return demand
 
-    def calc_route(self, impedance: dict, marine_impendance: dict, 
-                   origs: dict, dests: dict):
+    def calc_route(self, impedance: dict, origs: dict, dests: dict):
         """Calculates route choice for foreign freight trade. 
         
         Parameters
@@ -713,9 +712,6 @@ class FreightPurpose(Purpose):
         impedance : dict
             Freight assignment mode (truck/train/...) : dict
                 Type (time/dist/toll_cost/canal_cost) : numpy 2d matrix
-        marine_impedance : dict
-            Freight marine mode (container_ship/general_cargo...) : attribute
-                Type (dist/frequency) : numpy.ndarray
         origs : dict
             Origin border id (FIHEL/SESTO...) : str
                 Centroid id : int
@@ -724,9 +720,7 @@ class FreightPurpose(Purpose):
                 Centroid id : int
         
         """
-        imps = impedance.copy()
-        imps["ship"] = marine_impendance
-        costs = self.get_costs(imps, origs, dests)
+        costs = self.get_costs(impedance, origs, dests)
 
     def get_costs(self, impedance: dict, origs: dict = None, dests: dict = None):
         """Fetches calculated costs for each mode in model's mode choice.
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -724,7 +724,7 @@ class FreightPurpose(Purpose):
                 Centroid id : int
         
         """
-        imps = {key: impedance[key] for key in impedance if key in self.modes}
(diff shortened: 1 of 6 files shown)


Scripts/valma_freight.py
```diff
@@ -41,28 +42,37 @@ def main(args):
...
+        log.info(f"Calculating route for foreign purpose: {purpose.name}")
+        ship_imps, origs, dests = ass_model.freight_network.read_ship_impedances(
+            is_export=True)
+        purpose.calc_route(impedance, ship_imps, origs, dests)
+    
+    # prepare domestic model by splicing impedances and initializing final demand matrix 
+    for ass_class in list(impedance):
+        for mtx_type, mtx in impedance[ass_class].items():
+            impedance[ass_class][mtx_type] = mtx[:zonedata.nr_zones, :zonedata.nr_zones]
+    total_demand = {mode: numpy.zeros([zonedata.nr_zones, zonedata.nr_zones], dtype="float32")
```
Konnde: Might need to move this total demand to be initialized before trade route model later in order to add its demand. But within current development stage, I think it is better to keep this here now.

Scripts/valma_freight.py
```diff
@@ -41,27 +42,34 @@ def main(args):
...
                     for mode in param.truck_classes}
```
Konnde, line 70: Might need to initialize this before trade route model later so that we can sum that model's results here. But this fits our current logic so rather keep it here.

Scripts/datatypes/purpose.py
```diff
@@ -720,7 +724,9 @@ def calc_route(self, impedance: dict, origs: dict, dests: dict):
                 Centroid id : int
         
         """
-        costs = self.get_costs(impedance, origs, dests)
+        imps = impedance.copy()
```
zptro: Why do we need to copy this?

Scripts/valma_freight.py
```diff
@@ -41,27 +42,34 @@ def main(args):
...
-                        if mode in impedance[mtx_type]}
-                        for mode in ("truck", "freight_train", "ship")}
     
-    total_demand = {mode: numpy.zeros([zonedata.nr_zones, zonedata.nr_zones], 
-                                      dtype="float32")
+    # Run foreign trade route choice
+    for purpose in purposes.values():
+        log.info(f"Calculating route for foreign purpose: {purpose.name}")
+        ship_imps, origs, dests = ass_model.freight_network.read_ship_impedances(
+            is_export=True)
```
zptro: Why do you loop this?

Scripts/datatypes/purpose.py
Konnde: At this point our impedance includes truck classes, freight train and domestic ship. Without copying or making a new dict here, the domestic ship impedances gets overwritten with the marine ship impedances. The program then either crashes at impedance splicing for domestic model, or latest at calculating domestic ship costs. I moved this from valma freight so that we can create a temp copy to use for route impedance cost calc. After that it serves no purpose and garbage collector will clean it here instead of us needing to null it in valma_freight side. We can also look into other solutions if needed.

Scripts/valma_freight.py
Konnde: This reading marine impedances was here from previous development commits, and I didn't touch it because I felt it would not be in scope of this specific PR. I mean that I can move it out of the loop, because that is what I will do anyway later in the other upcoming PRs when I am building together this trade route model concept. But it is not specifically related to this PR.

Scripts/datatypes/purpose.py
zptro: Domestic ship and foreign ship do not seem to have _any_ shared features. Maybe you could simply use a different name? I think that would make `freight_costs.calc_cost` more simple as well.

Konnde: That is an option too. There is no reason to leave the foreign ship impedances to the impedance dict after trade route model has been calculated so we can delete this new "marine_ship" mode afterwards. I will make a version of this.

zptro (approved):
Looks really good!

PR #331, opened 2026-01-09
Remove extra assignment parameters

attesn:
Assignment parameters file has many unused parameters, so removing them for better readability. Should remove only deprecated features from code.


zptro (approved):
Thanks! 🙏

PR #332, opened 2026-01-09
Update emmebank dimensions

juliuskroetzl:
Increase submodel dimensions, because submodel ita_suomi (2018) has 308630 transit line segments


zptro:
@juliuskroetzl Did you check that also 2018 network fits?

juliuskroetzl:
> @juliuskroetzl Did you check that also 2018 network fits?

Yes, I specifically tested the 2018 network

zptro (approved):
Very good!

PR #334, opened 2026-01-09
Remove CBA calculation from model-system

attesn:
CBA calculation is being developed in other repository at the moment, so remove it from here to prevent parallel (extra) work and enhance code readability.


attesn:
@zptro

zptro (approved):

PR #336, opened 2026-01-12
Remove deprecated agent model

attesn:
Agent model is not currently in use or planned to be implemented before model release. Therefore I suggest that we remove deprecated agent model and related submodels. This helps model development in two ways:
1) Enhance code readability. Important especially for new users, who not only have to understand the working code, but the  unused material also.
2) Remove need to update deprecated features and tests. For example tour combinations was already replaced, but could not be removed as agent model tests would give unknown errors.

If agent model is later implemented we can use old datatypes as an example and build on that.


attesn:
@zptro

zptro (approved):

PR #338, opened 2026-01-13
Functionalize logistics module entry point

Konnde:
Move entry point from valma freight to a separate script as a function to:
- utilize same function in valma freight and unittest
- concept is easier to test
- reduce code in valma freight

Now it's initially inserted into freight_utils.py, and can be moved elsewhere if we have a better file location for this. As a heads up, similar entry point function concept will be added for foreign trade choice.


Scripts/utils/freight_utils.py
```diff
@@ -51,6 +51,52 @@ def create_purposes(parameters_path: Path, zonedata: FreightZoneData,
                                              resultdata, purpose_cost)
     return purposes
 
+def run_logistics_module(purpose: FreightPurpose, demand_truck : numpy.ndarray,
+                         impedance: numpy.ndarray, zonedata: FreightZoneData, 
+                         zone_index_map: dict, iterations: int) -> Tuple[numpy.ndarray]:
```
zptro: I would suggest placing this function either as a method in `FreightPurpose` or as a static function in `logistics` module.

zptro (changes requested):
Very good! Here are options for placing the function.

Scripts/utils/freight_utils.py
Konnde: FreightPurpose is actually good idea here as we have purpose as a function parameter so this can naturally slide as a method. Will follow this with updated suggestion.

Scripts/valma_freight.py
```diff
@@ -75,23 +74,10 @@ def main(args):
...
-                                                           lc_sizes=numpy.array(lcs_sizes.values))
-            for i in range(args.logistics_iterations):
-                final_demand = process_logistics_inference(model=logistics_module,
-                                                            n_zones=zonedata.nr_zones,
-                                                            demand=demand["truck"])
-                demand["truck"] = final_demand
+        if hasattr(purpose, "logistics_module") and args.logistics_iterations > 0:
+            demand["truck"], _ = purpose.run_logistics_module(demand["truck"], impedance, 
+                                                              ass_model.mapping, 
+                                                              args.logistics_iterations)
```
zptro, line 80: Now that the function is a method in `FreightPurpose`, the external check `hasattr` could be internalized as well.

zptro (approved):

Scripts/valma_freight.py
Konnde: Yeah that is true. However I am going to introduce new updates here later as I am dismantling the TradeRouteChoice branch so we can check this suggestion at that point too, as I think we have now fulfilled this pull request's main objective.

PR #339, opened 2026-01-15
Form impedance splitting concept

Konnde:
Impedance matrices for foreign freight trade route choice are masked into three different "splits" or "legs" that are
- Zones in Finland (including Finnish border control points) - Finnish border control points
- Finnish border control points - foreign border control points
- Foreign border control points - country cluster zones

The following order would be for when goods are exported, and in reverse when goods are imported. The read_ship_impedances function in freight_assignment.py handles classifying border control points into origins and destinations based on is_export. Rest are handled here.

The premise is that we handle mode alternatives, impedance matrices and their shapes here. And when later Trade Route Choice probability calculation is being implemented, we have already these parameters handled and can focus on utility functions etc.


Scripts/datatypes/purpose.py
```diff
@@ -705,23 +708,108 @@ def calc_traffic(self, impedance: dict):
         demand = {mode: (probs.pop(mode) * generation).T for mode in self.modes}
         return demand
 
-    def calc_route(self, impedance: dict, origs: dict, dests: dict):
-        """Calculates route choice for foreign freight trade. 
-        
+    def get_split_impedances(self, impedance: dict, origs: dict, dests: dict,
+                             is_export: bool) -> dict:
+        """Forms impedance matrices for the three route splits of foreign 
+        trade route choice model.
```
zptro: I would prefer "leg" instead of "split". At least I associate the word split with a split between alternatives in this context (e.g., modal split), while "legs" are more unambiguous.

Scripts/datatypes/purpose.py
```diff
@@ -705,23 +708,108 @@ def calc_traffic(self, impedance: dict):
...
+            Whether data should be fetched for export (True) or import (False)
+
+        Returns
+        -------
+        dict
+            Split name (1/2/3) : dict
+                Mode (truck/train/marine ships) : mask indexed numpy 2d matrix
         """
+        def _slice(matrix, row_mask, col_mask):
+            return matrix[numpy.ix_(row_mask, col_mask)]
```
zptro: If we want this as a separate function, maybe a lambda function would do? And please place it immediately before it is used the first time.

Scripts/datatypes/purpose.py
```diff
@@ -705,23 +708,108 @@ def calc_traffic(self, impedance: dict):
...
+        dest_zones = numpy.array(list(dests.values()), dtype=numpy.int32)
+        all_zones = self.generation_zone_data.all_zone_numbers
+        masks = {
+            "orig_borders": numpy.isin(all_zones, orig_zones),
+            "dest_borders": numpy.isin(all_zones, dest_zones),
+            "fin_zones": numpy.isin(all_zones, numpy.union1d(
+                                    self.orig_zone_numbers, orig_zones)),
+        }
+        masks["cluster_zones"] = ~masks["fin_zones"] & ~masks["dest_borders"]
+        return masks
```
zptro: You only call this method once, so why even have it a separate method? 

You are also not utilizing the dict features of the returned dict, but rather using the dict items as pseudo-variables. Why not create actual variables?

Scripts/datatypes/purpose.py
```diff
@@ -705,23 +708,108 @@ def calc_traffic(self, impedance: dict):
...
+        # Get costs and update impedance structure for marine ships
         costs = self.get_costs(impedance, origs, dests)
+        costs[marine_ships_name] = costs[marine_ships_name]["cost"]
+        for mode in costs[marine_ships_name]:
+            costs[marine_ships_name][mode].update(
+                {"frequency": impedance[marine_ships_name][mode]["frequency"]})
+        masks = self._form_split_masks(origs, dests)
+        
+        split_impedances = {
+            True: {
```
zptro: Could you please use a simple `if` - `else` structure instead of this temporal dict?

Scripts/datatypes/purpose.py
```diff
@@ -705,23 +708,108 @@ def calc_traffic(self, impedance: dict):
...
+                "split_one": {
+                    "truck": {
+                        "cost": _slice(costs["truck"]["cost"], masks["fin_zones"], 
+                                       masks["orig_borders"])
+                    },
+                    "freight_train": {
+                        "cost": _slice(costs["freight_train"]["cost"], masks["fin_zones"], 
+                                       masks["orig_borders"])
+                    }
```
zptro: Could you add these to the dict using for-loops over `("truck", "freight_train")`

zptro (changes requested):
Great to see constant progress! Here are some suggestions of how to make the code more concise.

Konnde:
Thanks for the comments and suggestions. I agree especially with the leg/split discussion, because personally I'm not content with current naming principle (as it is, like you suggest, misleading or vague) but I had to output something. I will address these.

zptro (approved):
Awesome!

PR #340, opened 2026-01-19
Calibrate assignment parameters for long distance transit

attesn:


attesn:
@jalovaj2 I added new branch for parameter changes. You can add further changes here.

PR #342, opened 2026-01-20
Change time interval for log file rotation

### LINKED ISSUE #341: Model run splits into multiple log files
https://github.com/Traficom/lem-model-system/blob/035fc2f5abb927caa0ed2c0c16dffabe507904b9/Scripts/utils/log.py#L42

Basic model runs can take now >48h, so one run splits into multiple log files, which makes them difficult to read. We should increase time, revert back to 'midnight' or just initialize new log when new model run is started.

attesn:
Revert file rotation to midnight as model run duration can be anything from 1h to 48h.  

Closes #341


zptro (approved):
This is much better!

PR #343, opened 2026-01-21
Raise error if long dist demand is called before it is calculated

attesn:
If daily and long dist demand forecasts are done same time, but in non expected order, the check for `demand_vrk.omx` does not work and error will happen only after 5-6 hours. So check that long dist model run is calculated before it is used as input data.


zptro (approved):
Very clever. Amazing that you realized what I have tried to achieve here, I struggle with that myself sometimes.

PR #345, opened 2026-01-22
Add inheritance to logistics module

Konnde:
Foreign trade route choice model is going to be integrated later into the **logistics.py** file. To make this integration easier and more explicit, I suggest adding _FreightDetourInference_ parent class with common attribute handling and few common methods. Both _LogisticsModule_ and later in another PR to be added _TradeRouteModule_ inherit from this class.

Each foreign trade model json is going to have similar "route_choice" parameter structure.


Scripts/models/logistics.py
```diff
@@ -79,81 +85,81 @@ def forward(self, probs_indices: np.ndarray) -> np.ndarray:
...
+            final_demand[origin_offset:origin_offset+current_batch_size, :] += direct
+            total_per_route += route_totals
+        
+        return origin_offset, current_batch_size
+
+def run_logistics_model(model: LogisticsModule, iteration: int) -> Tuple[np.ndarray]:
+        """Domestic logistics model entry point
+        Process full matrix in origin batches to limit memory usage
+        Process full matrix in origin batches in parallel
+        """
```
Konnde: At this point I decided to leave this run command as a function rather than method, and we can later test with a larger network and see if threading causes problems in case this was a run_model(self) method.

Scripts/datatypes/purpose.py
```diff
@@ -857,13 +848,9 @@ def run_logistics_module(self, demand_truck: numpy.ndarray, impedance: numpy.nda
...
-                                                       lc_indices=lc_indices,
-                                                       lc_sizes=lc_sizes)
+        cost = self.get_costs(impedance)
+        model = LogisticsModule(cost, self.route_params, lc_indices, lc_sizes)
         for i in range(1, iterations + 1):
-            demand_truck, per_route = process_logistics_inference(model=logistics_module,
-                                                                  demand=demand_truck,
-                                                                  iteration=i)
+            model.demand = demand_truck
+            demand_truck, per_route = run_logistics_model(model, i)
```
zptro: This is an improvement. However, I still think it is a bit bit strange to initialize `demand` and `cost` as object variables, instead of arguments to `run_logistics_model`.

zptro (dismissed):
Nice improvements!

Scripts/datatypes/purpose.py
Konnde: The demand I completely agree with as the function outputs demand. As demand later comes out altered, it makes sense that we give it in. So I will make this demand change commit.

The impedances on the other hand I feel are just static tools to use like the estimated model parameters, and therefore could give reasoning to lump them in with the constructor. But as we finish here, and I soon push TradeRouteModel PR, we can polish this structure so that it fits both.

zptro:
I think this looks good, but maybe we could let @jalovaj2 get the final word on this one?

jalovaj2 (approved):
Looks good! Good work especially including the batch processing inside the class since previously it was unnecessarily outside the class.

PR #346, opened 2026-01-23
Remove work-leisure segmentation from assignment classes

attesn:
Pros: 
* Reduce running times (koko Suomi: 1,5h for free flow and 8h for model with peak hours).
* Reduce storage needed for OMX matrices (koko Suomi: 9,28 GB => 7,72 GB for free flow).
* Cleaner user interface, as there are less assignment classes to account in analysis.

Neutral:
* Mode and destination choice does not change.
* Park and ride modes are anyways either work or leisure.

Cons:
* Route choice between rail and bus does not account for different VoT. However, effect in assignment was small. Our validation data cannot show any difference to better or worse.
* CBA is not as accurate. This could be compensated by creating CBA end-assignment that could have more detailed segmentation (work, leisure, business).


zptro (dismissed):
No entirely convinced about the necessity, but I will approve this for now, and if needed we will get back to this later

attesn:
Re-review due to recent updates of `dev` branch.

zptro (approved):

PR #347, opened 2026-01-26
Export transit demand from long distance model

attesn:
This PR fixes two related errors:
1) Export transit matrices from long distance run by using `simple_transit_classes` in place of `long_dist_simple_classes`.
2) Add `transit_work` and `transit_leisure` demand shares for external traffic import.


attesn:
Had misunderstanding of how external demand works. Update upcoming in few minutes.

zptro (approved):

PR #348, opened 2026-01-26
Remove reference to car driver share parameter

attesn:
If-else statement not in use. Reference parameter removed in #336.


zptro (approved):
The functionality has been changed a long time ago, so this artifact is good to remove.

PR #349, opened 2026-01-29
Implement foreign trade entry point

Konnde:
Add similar entry point style as logistics module to run foreign trade route choice module. While the main focus here was extending FreightPurpose and logistics.py with the run tools, I also did add draft of is_export conditions.


Scripts/datahandling/matrixdata.py
```diff
@@ -138,3 +138,15 @@ def mapping(self, zone_numbers):
     @property
     def matrix_list(self):
         return self._file.list_matrices()
+
+def read_omx_item(path: Path, key: str):
+    with omx.open_file(path) as omx_file:
+        try:
+            demand = numpy.array(omx_file[key])
```
Konnde: Added this here because can't read in trade demand through MatrixFile's getter since its been designed for full matrices and these ones are not.

Scripts/datatypes/purpose.py
```diff
@@ -688,6 +689,27 @@ def __init__(self, specification, zone_data, resultdata, costdata):
         
         self.modes = list(specification["mode_choice"])
         self.route_params = specification.get("route_choice", None)
+        self.is_export = {"export": True, "import": False}.get(specification["struct"])
+        self._check_variables(specification)
+
+    def _check_variables(self, specification):
+        msg = ""
+        if len(self.modes) == 0:
+            msg = "No modes in specification"
```
Konnde: I kept making typos in the I/O while I was designing this feature so I added these here for the time being. Can remove altogether from this PR but would also be nice if we had a better place to validate I/O conditions.

Scripts/datatypes/purpose.py
```diff
@@ -853,3 +874,56 @@ def run_logistics_module(self, demand_truck: numpy.ndarray, impedance: numpy.nda
...
+                zones = numpy.array(list(trade_mappings[mapping]))
+                if mapping == "finland_zone_number":
+                    row_mask = numpy.isin(zones, all_zones)
+                elif mapping == "cluster_zone_number":
+                    col_mask = numpy.isin(zones, all_zones)
+                else:
+                    msg = f"Invalid mapping '{mapping}'"
+                    log.warn(msg)
+                    raise ValueError(msg)
+            demand = demand[row_mask][:, col_mask]
```
Konnde: Very rough draft just for now and would like suggestions here how we could best add municipality center 
mapping integration. While mainly this tool is planned to be run with koko suomi, and for testing we can just make up demand mtx of specific shape, speed and testing-wise it would be beneficial to also be able to run this section of the model with municipality center mapping.

Scripts/models/logistics.py
```diff
@@ -163,3 +185,22 @@ def run_logistics_model(model: LogisticsModule, demand: np.ndarray,
         f"orig_demand {np.sum(demand)}, final_demand {np.sum(final_demand)}"
     )
     return final_demand, total_per_route
+
+def run_trade_model(model: TradeRouteModule, demand: np.ndarray):
+    """Using given truck demand full matrix, calculates share of demand for 
+    mode-route alternatives between Finnish zones and foreign country clusters
+    through designated border control points.
```
Konnde, line 192: Implement this feature in next PR.

Konnde (commented):
Few notes here which I deemed most important, and I'm up for new suggestions.

Scripts/datatypes/purpose.py
```diff
@@ -688,6 +689,27 @@ def __init__(self, specification, zone_data, resultdata, costdata):
         
         self.modes = list(specification["mode_choice"])
         self.route_params = specification.get("route_choice", None)
+        self.is_export = {"export": True, "import": False}.get(specification["struct"])
```
zptro, line 690: Could these be rewritten so that they fail if the specification is invalid? I mean instead of the separate variable check?

Scripts/valma_freight.py
```diff
@@ -52,15 +52,22 @@ def main(args):
...
+            ship_imps, origs, dests = marine_export[0], marine_export[1], marine_export[2]
+        else:
+            ship_imps, origs, dests = marine_import[0], marine_import[1], marine_import[2]
+        demand, trade_mappings = read_omx_item(Path(args.trade_demand_data_path), purpose.name)
         impedance[param.marine_ships_name] = ship_imps
-        impedance_legs = purpose.form_impedance_legs(impedance, origs, dests, is_export)
+        impedance_legs = purpose.form_impedance_legs(impedance, origs, dests)
+        demand = purpose.run_trade_route_module(impedance_legs, origs, dests,
+                                                demand, trade_mappings)
     del impedance[param.marine_ships_name]
```
zptro: I suggest moving all of this into `purpose.run_trade_route_module()`

Scripts/datatypes/purpose.py
zptro: Should this be `demand = demand[row_mask, col_mask]`?

Scripts/datahandling/matrixdata.py
```diff
@@ -138,3 +138,15 @@ def mapping(self, zone_numbers):
...
+        except NodeError:
+            msg = f"Invalid omx key '{key}' in {path.name}"
+            log.error(msg)
+            raise NodeError(msg)
+        mappings = {name: omx_file.mapping(name)
+                    for name in omx_file.list_mappings()}
```
zptro: `mapping` is a wrapper around `map_entries` method that creates a lookup table map -> offset. What you seem to need however, is the simple list returned by `map_entries`.

Scripts/datatypes/purpose.py
Konnde: I will rework this little bit. Personally, I have noticed that I'm not fond of having error and conditional checks in the constructor because I feel its purpose is to just present a clear "window" on what variables each object has. Other included design choices (like the the check_variable method here) in constructor hinder this purpose.

Scripts/datatypes/purpose.py
Konnde: As far as I've understood, when trying to do matrix[row splicing, col splicing] with either of the given splices being 1d arrays, numpy defaults to giving resulting matrix as 1d. Here both splices are 1d bool arrays, so chained indexing is needed to get both dimensions.

zptro: I see. I think what you are looking for is `demand = demand[numpy.ix_(row_mask, col_mask)]`

zptro: I have mostly used index interval slicing in other places, as we have the centroids in a certain order (domestic, foreign, border points in separate index intervals), as using boolean masks is generally more difficult to read. I admit `numpy._ix()` is not very self-explanatory, but at least it is a fairly commonly used function.

Scripts/parameters/freight/foreign/kemlaa_import.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "kemlaa_import",
+    "orig": "null",
+    "dest": "null",
+    "generation_area": "foreign",
+    "attraction_area": "foreign",
```
zptro: We need to have either of these as "domestic", to be able to access domestic zone data.

Konnde: We can have this. If specification is for export, we can do generation_area as "domestic", and for importing we can have attraction area as "domestic". In both of these cases we would then be referring to Finland as domestic and abroad as foreign.

Scripts/datatypes/purpose.py
```diff
@@ -853,3 +874,56 @@ def run_logistics_module(self, demand_truck: numpy.ndarray, impedance: numpy.nda
...
+            for mapping in trade_mappings:
+                zones = numpy.array(list(trade_mappings[mapping]))
+                if mapping == "finland_zone_number":
+                    row_mask = numpy.isin(zones, all_zones)
+                elif mapping == "cluster_zone_number":
+                    col_mask = numpy.isin(zones, all_zones)
+                else:
+                    msg = f"Invalid mapping '{mapping}'"
+                    log.warn(msg)
+                    raise ValueError(msg)
```
zptro: Something like this should be sufficient

```suggestion
        if self.generation_zone_data.mapping.name == "municipality_center":
            df = pandas.DataFrame(demand, trade_mappings["finland_zone_number"])
            demand = df.groupby(self.generation_zone_data.mapping).sum().to_numpy()
```

Scripts/datatypes/purpose.py
```diff
@@ -853,3 +864,43 @@ def run_logistics_module(self, demand_truck: numpy.ndarray, impedance: numpy.nda
...
+            _description_
+        """
+        impedance_legs = self.form_impedance_legs(impedance, origs, dests)
+        demand, trade_mappings = read_omx_item(trade_demand_path, self.name)
+
+        mapping_name = (self.generation_zone_data.mapping.name if self.is_export 
+                        else self.attraction_zone_data.mapping.name)
+        if mapping_name == "municipality_center":
+            df = pandas.DataFrame(demand, trade_mappings["finland_zone_number"])
+            demand = df.groupby(self.generation_zone_data.mapping).sum().to_numpy()
```
Konnde, line 905: Tested with given design suggestion in municipality network and got through with correct shape, great!

Scripts/valma_freight.py
```diff
@@ -56,18 +56,16 @@ def main(args):
...
     marine_import = ass_model.freight_network.read_ship_impedances(False)
     for purpose in purposes.values():
-        log.info(f"Calculating route for foreign purpose: {purpose.name}")
+        log.info(f"Calculating trade routes for foreign purpose: {purpose.name}")
         if purpose.is_export:
             ship_imps, origs, dests = marine_export[0], marine_export[1], marine_export[2]
         else:
             ship_imps, origs, dests = marine_import[0], marine_import[1], marine_import[2]
-        demand, trade_mappings = read_omx_item(Path(args.trade_demand_data_path), purpose.name)
         impedance[param.marine_ships_name] = ship_imps
```
zptro: Slowly grasping the whole picture. I would suggest feeding `ship_imps` to `purpose.run_trade_route_module` as a separate argument instead of appending it to and later deleting it from `impedance`. I mean, they are added to `impedance_legs` in a completely independent step, so `purpose` could just as well call `freight_costs.get_foreign_ship_cost` directly instead of digging it out of `costs` using an external key.

Scripts/datatypes/purpose.py
```diff
@@ -853,3 +864,43 @@ def run_logistics_module(self, demand_truck: numpy.ndarray, impedance: numpy.nda
...
+
+        fin_port_indices = origs if self.is_export else dests
+        port_names = dict(zip(fin_port_indices.keys(), range(len(fin_port_indices))))
+        port_indices = numpy.array(tuple(port_names.values()))
```
zptro: Would this be a bit clearer?

```suggestion
        port_indices = numpy.arange(len(fin_port_indices))
```

zptro (commented):
Really good! Here are some detailed thoughts.

Scripts/datatypes/purpose.py
Konnde: We could do that here. port_names' significance is not shown in this PR but in the next one, so I can just add it back at that point

Scripts/valma_freight.py
Konnde: I personally think it is more explicit to add the marine ship impedances (be it within valma_freight or run_trade_model) with the others and then calculate costs for the hash map container rather than first calc costs for road and rail and then separately for marine ships. But I can make the changes here if desired.

It means we break away from the design style of calc_cost function being our entry point to getting any costs and thus should then remove mentions of marine_ships from there too. At that point we could remove mention of marine_ships altogether from the commodity specifications too as they're not referred anywhere and users just have to know that the marine ships are added during run separately from truck and freight_train.

Scripts/valma_freight.py
```diff
@@ -50,21 +51,18 @@ def main(args):
...
+            demand = purpose.run_trade_route_module(impedance, *marine_export,
+                                                    trade_demand_path)
         else:
-            ship_imps, origs, dests = marine_import[0], marine_import[1], marine_import[2]
-        impedance[param.marine_ships_name] = ship_imps
-        demand = purpose.run_trade_route_module(impedance, origs, dests,
-                                                Path(args.trade_demand_data_path))
-    del impedance[param.marine_ships_name]
+            demand = purpose.run_trade_route_module(impedance, *marine_import,
+                                                    trade_demand_path)
```
Konnde, line 65: The change does allow us to unpack cleanly into running trade model like this without needing to create extra variables 👍

zptro (approved):
Looks good!

PR #350, opened 2026-02-04
Add missing return trips to car egress modes

attesn:


zptro (approved):
Much nicer solution, thanks!

PR #352, opened 2026-02-04
Change draught check and impedance leg formation

zptro:
- Tried to make the draught check more simple and easy to follow
- Tried to make impedance leg formation easier to maintain by generalizing and removing duplicate use of key words
- Removed origs/dests border id logic, as handling is mostly different for Finland and foreign border ids

### DIFF
--- a/Scripts/assignment/freight_assignment.py
+++ b/Scripts/assignment/freight_assignment.py
@@ -147,10 +147,10 @@ class FreightAssignmentPeriod(AssignmentPeriod):
             Mode (container_ship/general_cargo...) : attribute
                 Type (dist/frequency) : numpy.ndarray
         dict
-            Origin border id (FIHEL/SESTO...) : str
+            Finland border id (FIHEL/FISKV...) : str
                 Centroid id : int
         dict
-            Destination border id (FIHEL/SESTO...) : str
+            Foreign border id (AEJEA/SESTO...) : str
                 Centroid id : int
         """
         fin_borders = self._filter_border_points(finland_border_points)
--- a/Scripts/assignment/freight_assignment.py
+++ b/Scripts/assignment/freight_assignment.py
@@ -153,26 +153,23 @@ class FreightAssignmentPeriod(AssignmentPeriod):
             Destination border id (FIHEL/SESTO...) : str
                 Centroid id : int
         """
-        if is_export:
-            origins = finland_border_points
-            destinations = cluster_border_points
-        else:
-            origins = cluster_border_points
-            destinations = finland_border_points
-        origins = self._filter_border_points(origins)
-        destinations = self._filter_border_points(destinations)
-        orig_mapping = {pid: idx for idx, pid in enumerate(origins)}
-        dest_mapping = {pid: idx for idx, pid in enumerate(destinations)}
+        fin_borders = self._filter_border_points(finland_border_points)
+        cluster_borders = self._filter_border_points(cluster_border_points)
+        fin_mapping = {pid: idx for idx, pid in enumerate(fin_borders)}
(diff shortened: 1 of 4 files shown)


Scripts/datatypes/purpose.py
```diff
@@ -904,8 +898,7 @@ def run_trade_route_module(self, impedance: dict, ship_imps: dict,
             df = pandas.DataFrame(demand, trade_mappings["finland_zone_number"])
             demand = df.groupby(self.generation_zone_data.mapping).sum().to_numpy()
 
-        fin_port_indices = origs if self.is_export else dests
-        port_indices = numpy.arange(len(fin_port_indices))
```
Konnde, line 908: Just a note that these should actually be the real indeces of the border control points similar to what domestic logistics model has and not just aranged. However I need to do small updates here anyway in next PR so we can leave this be.

Scripts/utils/freight_costs.py
```diff
@@ -193,8 +183,11 @@ def get_aux_cost(unit_costs: Dict[str, Dict],
         calc_road_cost(unit_costs, impedance_aux, model_category))
     return aux_cost
 
-def get_foreign_ship_cost(unit_costs: Dict[str, dict], impedance: Dict[str, dict], 
-                          model_category: str, origs: dict, dests: dict):
+def get_foreign_ship_cost(unit_costs: Dict[str, dict],
```
Konnde, line 186: Many clever updates in this function. We lose giving some explicit information to reader from crunching port draught evaluation into this function (as the current design has separated cost calc and separated draught eval), but I'd say that is a good trade here.

Scripts/utils/freight_costs.py
```diff
@@ -144,16 +136,14 @@ def calc_ship_cost(unit_costs: Dict[str, float],
...
     
     Returns
     -------
     ship_cost : numpy.ndarray
         impedance type cost : numpy 2d matrix
     """
-    ship_cost = (impedance["time"] * unit_costs["time"] * draught_mask
-                 + impedance["dist"] * unit_costs["dist"] * draught_mask
+    ship_cost = (impedance["time"] * unit_costs["time"]
+                 + impedance["dist"] * unit_costs["dist"]
```
Konnde, line 146: Happy that you were able to get rid of the draught mask here.

Scripts/assignment/freight_assignment.py
```diff
@@ -147,32 +147,29 @@ def read_ship_impedances(self, is_export: bool) -> Tuple[dict, dict, dict]:
...
         ship_impedances = {}
         for marine_ship, modes in param.freight_marine_modes.items():
             ship_impedances[marine_ship] = {}
             ship_mode = next(iter(modes))
             for key, attr in param.ship_attrs.items():
                 ship_impedances[marine_ship][key] = self._line_data_to_matrix(
-                    transit_lines[ship_mode], attr, orig_mapping, dest_mapping)
-        return ship_impedances, origins, destinations
+                    transit_lines[ship_mode], attr, *mappings)
+        return ship_impedances, fin_borders, cluster_borders
```
Konnde, line 172: Surprisingly I think this is the biggest change here as I have to update my mental thinking that we're not dealing with changing origins and destinations anymore but with static Finnish and Cluster border sequences. I do think it is better that they're static.

Konnde (dismissed):
Nice updates! I won't go over much as the general functionality is unchanged while there are logic and semantic improvements. Did some testing with municipality network and everything was working well.

Scripts/utils/freight_costs.py
zptro: Added comment to clarify

Konnde (approved):

PR #354, opened 2026-02-04
Move parking cost reweight to `transform_impedance` method

attesn:
Move parking cost reweight to `transform_impedance` method. All impedance transformations are in one place and extra instance method removed.

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -184,7 +184,7 @@ class Purpose:
                                                     cost.car_pax_occupancy[self.name])
                     except KeyError:
                         pass
-                airpl_access_modes = ["airpl_car_acc", "airpl_taxi_acc", "airpl_car_egr"]
+                airpl_access_modes = ["airpl_car_acc", "airpl_car_egr"]
                 transit_access_modes = ["pt_car_acc", "pt_taxi_acc", "pt_taxi_egr"]
                 if mtx_type == "time" and mode in airpl_access_modes:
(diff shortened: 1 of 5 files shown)


attesn:
After some thinking individual dummies do have role in choice model, so reverted two commits.

Scripts/datatypes/purpose.py
```diff
@@ -424,24 +428,15 @@ def calc_prob(self, impedance, is_last_iteration):
                     prob[acc_mode] = split[acc_mode] * main_prob
         return prob
 
-    def split_connection_mode(self, impedance, pt_mode, car_acc_modes):
-        access_modes = car_acc_modes + [pt_mode]
+    def split_connection_mode(self, impedance, pt_mode):
         if pt_mode == "airplane":
-            access_modes.append("airpl_taxi_acc")
```
zptro, line 430: This needs to be reverted also, right?

zptro (commented):
Very good, much clearer structure!

Scripts/datatypes/purpose.py
attesn: I think this is not needed as we take copy of all impedances, which now include reweighted parking costs. This parking cost is same for `airpl_car_acc` and `airpl_taxi_acc`. Or is there something that I missed?

zptro: Yes, you are right, this is part of what you refactored!

zptro (approved):

PR #355, opened 2026-02-06
Remove area based assignment results

attesn:
* Simplification removes one attribute to configure between network-builder and model-system. Also enables easier handling of aggregations. 
* Area based analyses are now easy to perform as "assignment_results.gpkg" is exported from model run.


attesn:
This PR would help to later handle `aggregate_results` as optional zonedata variable.

zptro (approved):
I guess this makes sense.

Regarding the area aggregations of demand, there is currently no other way to compare matrix-level results per purpose. Do you think we should remove that possibility altogether or do you have another solution in mind?

attesn:
> 
> Regarding the area aggregations of demand, there is currently no other way to compare matrix-level results per purpose. Do you think we should remove that possibility altogether or do you have another solution in mind?

My idea is that we would separate variables that need to be defined (area based dummy) and optional variables (if user wants to add some aggregation to results). For example `ykr_km` should not raise errors as it was added only to produce matrix results and model would work ok without it.

PR #357, opened 2026-02-09
Separate required and optional area aggregations

attesn:
Attempt to clarify area aggregation usage:
* Demand model dummies (municipality, county, submodel) must be found in zone data.
* Optional variables to aggregate results can be added to data. These are identified with `aggregate_results` prefix.  
* Rename "subarea" to "calibration_area" to highlight inteded use as calibration dummy.
* Remove currently unused dummy variables.


Scripts/datahandling/zonedata.py
```diff
@@ -363,14 +358,17 @@ def read_zonedata(path: Path,
             try:
                 total = col["total"]
             except TypeError:
-                aggs[col] = func
+                aggs[col] = "first"
```
zptro: I think you are dropping the "mean" aggregation type here. I guess that is not intentional?

zptro (commented):
Looks very good! Only one thing that I was thinking about.

Scripts/datahandling/zonedata.py
```diff
@@ -363,7 +358,10 @@ def read_zonedata(path: Path,
             try:
                 total = col["total"]
             except TypeError:
-                aggs[col] = func
+                if col.startswith("aggregate_results_"):
+                    aggs[col] = "first"
+                else:
+                    aggs[col] = func
```
zptro: I do not understand the need of this special case. If the aggregation type is "first", the column should be under "first" in the json file, right?

attesn: My mistake. Wrote different type of solution at first, which included this change.

zptro (dismissed):
Looks very good!

attesn:
@jmutikainen Can you adjust changes to `prepare-zone-data` repo? Meaning that we'll drop `aggregate_results` prefix from municipalities, counties and submodel.

attesn:
@zptro Re-review because I had to merge dev, as test zonedata was changed in #306.

zptro (approved):

PR #358, opened 2026-02-11
Move truck value-of-time to json input, change input of vot_inv to vot

zptro:
- Change format of assignment value-of-time input from inverted [min/eur] to direct [eur/hour]
- Separate car and transit VoT and make values explicitly assignment-class specific (this makes `parameters.assignment.assignment_classes` dict superfluous at least in assignment, might be moved to demand model json)
- Move truck VoT to cost json
- Update values according to latest Väylä guidelines, mostly using short-distance values, but for pure long-distance transit modes using long-distance values

@attesn I would not mind moving these assignment VoT values to `parameters.cost`, but should we still have separate values for the long-distance tour purposes?


attesn:
This is good update! I think it would be good to have only one VoT values in model-system and merge these into `parameters.cost`.

Konnde (dismissed):
Yep, looks good.

zptro:
@attesn would this be ok to merge now?

attesn (approved):

PR #359, opened 2026-02-16
Remove Excel export of aggregated matrices

attesn:
* There is dependency problem with `openpyxl` package. Package is now used only to print aggregated demand matrices. 
* Current format of matrices work well with Pivot Table, so direct Excel export can be deprecated. 
* Remove `openpyxl` from pipfile and requirements.txt, but keep requirements file as we might need it later.


attesn:
Should we export CSV files, so Windows would suggest opening aggregated demand matrices with Excel?

zptro (approved):

PR #360, opened 2026-02-16
Vehicle type based in-vehicle time weight

attesn:
Use separate weights for local and long distance trains. This PR requires changes to Emme-network:

New vehicle types: 
```
a 11 'vr_lahi' j      1   100   150   0.00   0.00   0.00   0.00   1.00
a 12 'kiskobussi' j      1   100   150   0.00   0.00   0.00   0.00   1.00
```

Following lines need to have vehicle type set to `11`: 

* Lahti-Kouvola-Kotka (O)
* Lahti-Helsinki (Z)
* Helsinki-Riihimäki-Nokia (R)
* Helsinki-Hameenlinna (D)
* Lahti-Riihimaki (G)
* Helsinki-Hanko (H)

Following lines need to have vehicle type set to `12`: 
* Savonlinna–Parikkala
* Pieksämäki–Joensuu
* Joensuu–Nurmes
* Tampere–Haapamäki–Jyväskylä
* Tampere–Keuruu
* Jyväskylä–Seinäjoki
* Seinäjoki–Ähtäri
* "Pandalino" Seinäjoki–Alavus.


Scripts/parameters/assignment.py
```diff
@@ -161,7 +161,9 @@
     'W': 0
 }
 in_vehicle_weight = {
-    'j': 0.6,
+    '3': 0.6,
+    '11': 0.8,
+    '12': 0.8
 }
```
jalovaj2, line 210: For Helsinki-Lahti and Lahti-Kouvola legs changing this as

```suggestion
in_vehicle_weight = {
    '1': 1, # Bus
    '2': 0.8, # Tram
    '3': 0.8, # Long-distance train
    '4': 0.8, # Metro
    '5': 1, # Ferry
    '6': 1, # Airplane
    '7': 0.8, # Light rail
    '8': 1, # Long distance bus
    '9': 1, # Local train
    '10': 1, # Trunk bus
    '11': 1, # Regional train
    '12': 1, # Railbus
    '13': 1, # Long-distance day Ferry
    '14': 1, # Long-distance night Ferry
}
```

gives significantly lower train traffic amounts which are clearly closer to the validation data. This is partly due long-distance train in_vehicle_weight changed from 0.6 to 0.8 before merging this PR, but the overall effect still seems better for these legs using the above parameters. However, for Helsinki-Hämeenlinna leg which previously was close to validation data is now clearly lower than validation data. Also, the legs that previously had lower train traffic than validation, such as Helsinki-Turku and Tampere-Pori now have even lower train traffic amounts.

jalovaj2 (changes requested):
Looks good! We still have to consider the parameter values regarding the local differences in train traffic amounts.

Scripts/parameters/assignment.py
attesn: Thanks for comprehensive report! Values now updated.

jalovaj2 (approved):
Looks good!

PR #361, opened 2026-02-18
Add dedicated model run for car fleet analysis

attesn:
Add dedicated car fleet model run:
* Segmented calculation and results for car fleet analysis. 
* No need to wait for whole model-system run, as car ownership is calculated at beginning. 

Work to do:
* I think best way to use this is post-model-run-analysis similar to CBA. However, here we need logit model methods, so it cannot be placed in other repo.
* Car type choice model possibly integrated later, depending of how complicated calculation will be.


attesn:
@zptro Draft so far, but ideas for further development are well appreciated.

zptro (approved):

rviri1 (approved):

PR #362, opened 2026-02-24
Implement trade route probability calc

Konnde:
Add TradeRouteModule in which import and export specific commodity demand flows are calculated using MNL model where demand from origin to destination passes through two sets of border control points (within Finland, outside Finland) to produce leg and mode specific demand matrices.

The design is inspired by LogisticsModule design choices but there are major differences including:
- Need to enable multiple mode alternatives on each leg (LogisticsModule only has truck)
- Need to pass through multiple border control points rather than just one set of logistics center
- Demand is not full matrix and the shape changes based on commodity's trade type (export/import)
- Commodity specific model parameter structures have much variation

As a design note, I have tried to design the structure so that we can enable multiple modes on all legs. This means that when exporting from Finland, we can have multiple mode alternatives on legs 1 and 2, and while importing, on legs 2 and 3. However I have understood that the estimations for Finland's land leg have been mainly done with just truck mode due to scarce rail related export and import trade data. This most likely means that we end up dropping freight_train as a mode alternative for  now and rather use domestic model's commodity specific mode shares.


Konnde:
After finishing the draft for TradeRouteModule, my intention was to open the logic for peer-review and to meanwhile work on the export and import related commodity json files. I had not realized that many of them have significant differences in utility calculation, and their structures would not pass in this current design, so I decided to close this PR for the time being. However, I think the decision was too hasty, as the general TradeRouteModel logic still stays intact and can be reviewed even if I have to do more updates around utility calculations. So I will reopen this PR so that the logic of TradeRouteModule can be reviewed while I continue to work on the commodity specific json files and later push the necessary changes.

Scripts/datatypes/purpose.py
```diff
@@ -887,15 +887,25 @@ def run_trade_route_module(self, impedance: dict,
...
-        port_indices = numpy.arange(len(fin_border_ids))
-
-        route_model = TradeRouteModule(impedance_legs, self.route_params, port_indices)
+        elif demand.shape[0] != self.orig_zone_numbers.size:
+            idx_names = ["finland_zone_number", "cluster_zone_number"]
+            idx = [None, None]
+            for i, name in enumerate(idx_names):
+                mapping_arr = numpy.array(trade_mappings[name], dtype=all_zones.dtype)
+                mask = numpy.isin(all_zones, mapping_arr)
+                idx[i] = numpy.where(mask)[0].astype(numpy.int32)
```
zptro: I think this one-liner would output the same thing.

```suggestion
                idx[i] = numpy.searchsorted(all_zones, trade_mappings[name]).astype(numpy.int32)
```

Konnde:
After discussing with @zptro about splicing operations that might be unnecessary in the calculation logic, I further looked into each calculation step and indeed found several pointless operations that I had left in which we could drop to further simplify code structure and logic. In addition, added trade related commodity model specifications that have been so far estimated. There are few remaining commodities that don't yet have satisfied estimation results, and they might or not get added within this PR.

Scripts/parameters/freight/foreign/elektr_export.json
```diff
@@ -0,0 +1,77 @@
...
+            },
+            "container_ship": {
+                "first_class": [
+                    2.295,
+                    ["FIRAU", "FIHMN", "FIHEL"]
+                ],
+                "second_class": [
+                    2.634,
+                    ["FIOUL", "FIKEM"]
+                ]
```
zptro: So these class-name keys are not used anywhere, right? Would it be more explicit to have the port names as keys directly (e.g., `"FIRAU FIHMN FIHEL": 2.295`)? At least I think it would make `add_constants` easier to understand.

Scripts/models/logistics.py
```diff
@@ -132,8 +122,158 @@ def process_batch(self, origin_offset: int, n: int, k_plus1: int, demand: np.nda
...
+        if mode in self.constant:
+            for class_items in self.constant[mode].values():
+                borders = set(self.fin_borders) & set(class_items[1])
+                border_constants.update({border: class_items[0] for border in borders})
+        for border, constant in border_constants.items():
+            idx = self.fin_borders[border]
+            if self.is_export:
+                constants_mtx[idx, :] += constant
+            else:
+                constants_mtx[:, idx] += constant
```
zptro: I think this could be written as two nested for-loops without using temp dict `border_constants`.

Scripts/models/logistics.py
```diff
@@ -132,8 +122,158 @@ def process_batch(self, origin_offset: int, n: int, k_plus1: int, demand: np.nda
...
     def __init__(self, impedance: Dict[str, np.ndarray], model_parameters: dict, 
-                 port_indices: np.ndarray):
-        FreightDetourInference.__init__(self, impedance, model_parameters, port_indices)
+                 fin_borders: dict, is_export: bool):
+        FreightDetourInference.__init__(self, impedance, model_parameters)
+        self.fin_borders = fin_borders
+        self.is_export = is_export
+        self.leg1_modes = list(impedance["leg_one"])
+        self.leg2_modes = list(impedance["leg_two"])
+        self.leg3_modes = list(impedance["leg_three"])
```
zptro: My main issue with this class is the mix of hard-coded attribute names and corresponding dict keys. For instance, these hard-coded attributes seem a bit unnecessary, because in the places where you use them, you also use the corresponding dict keys for referring to other stuff.

zptro (commented):
Nice to see that you were able to simplify the previous draft. Here are some further comments.

Scripts/parameters/freight/foreign/elektr_export.json
Konnde: Currently they're implications of the estimated ship mode's service class and can be dropped if there's not need for more explicit design.

I personally don't think "container_ship": "FIRAU FIHMN FIHEL": 2.295 is any better than nested list "container_ship": [2.295, ["FIRAU", "FIHMN", "FIHEL"]] because the key has to be split anyway and there are specifications where keys can become quite long

"FIUKI FITKU FIRAA FIPOR FIKOK FIKEM FIKAS FIINK FIRAU FIHMN FIHEL": 1.0

Writing each border point: constant separately would be most efficient design in code but the specification would have a lot of value duplicates

"FIRAU": 2.295,
"FIHMN": 2.295 ....

Scripts/models/logistics.py
Konnde: There are hard coded values yes but they're either model specification related or leg name related.

The leg name related hard coding I can get rid of by adding the leg names into parameters file and reference here.

Scripts/parameters/freight/foreign/elektr_export.json
zptro: In that case, could it rather be a dict? For instance:

```
"first_class": {
    "border_points": "["FIRAU", "FIHMN", "FIHEL"],
    "dummy": 2.295
},
```

Konnde: I think this could be the best middle ground. I'll do test updates for the two kemlaa test commodities.

Scripts/models/logistics.py
```diff
@@ -99,10 +89,11 @@ def process_batch(self, origin_offset: int, n: int, k_plus1: int, demand: np.nda
                                    + current_batch_size, dtype=np.int32)
         dest_indices = np.arange(n, dtype=np.int32)
         o_grid, d_grid = np.meshgrid(origin_indices, dest_indices, indexing='ij')
-        eval_indices = np.stack([o_grid.ravel(), d_grid.ravel()], axis=1)
+        origin_indices = o_grid.ravel()
+        destination_indices = d_grid.ravel()
 
         # Get probabilities and reshape
-        probs_batch_flat = self.forward(probs_indices=eval_indices)
+        probs_batch_flat = self.forward(origin_indices, destination_indices)
```
zptro, line 96: This is so much clearer! 👍

Scripts/models/logistics.py
```diff
@@ -132,9 +123,131 @@ def process_batch(self, origin_offset: int, n: int, k_plus1: int, demand: np.nda
...
+        route_util = (
+            util_one[:, :, None, None]
+            + util_two[None, :, :, None]
+            + util_three[None, None, :, :]
+        )
 
+        n_origin, n_orig_alt, n_dest_bcp_alt, n_dest = route_util.shape
+        route_flat = route_util.reshape(n_origin, n_orig_alt * n_dest_bcp_alt, n_dest)
+        route_probs_flat = self.softmax(route_flat, axis=1)
+        route_probs = route_probs_flat.reshape(n_origin, n_orig_alt, n_dest_bcp_alt, n_dest)
```
zptro:
```suggestion
        route_probs = route_probs_flat.reshape(route_util.shape)
```

Scripts/models/logistics.py
```diff
@@ -132,9 +123,131 @@ def process_batch(self, origin_offset: int, n: int, k_plus1: int, demand: np.nda
...
+        prob_orig_bcp = np.moveaxis(prob_orig_bcp, 2, 1)
+        prob_dest_bcp = np.moveaxis(prob_dest_bcp, 3, 1)
+        demand_routed = (
+            demand[:, :, None, None]
+            * prob_orig_bcp[:, :, :, None]
+            * prob_dest_bcp
+        )
+        
+        # axes now = (n_orig (batched), n_dest, n_orig_alt, n_dest_bcp_alt)
+        # keep axes defining each leg segment, sum out the rest
```
zptro: Not batched though?

Scripts/models/logistics.py
```diff
@@ -187,20 +300,45 @@ def run_logistics_model(model: LogisticsModule, demand: np.ndarray,
...
     Returns
     -------
-    __type__
-        _description_
+    dict
+        Leg name (one/two/three) : Mode
+            Name (truck/container_ship...) : numpy 2d array
+    """
+    len_leg_two = len(model.impedance.get("leg_two"))
+    flows = model.route_demand(demand, len_leg_two)
```
zptro: You are using `len_leg_two` only in `model`, so maybe you could move the definition of it inside `model`?

zptro (commented):
Wow, this looks really good now! Just some minor comments left.

Scripts/models/logistics.py
Konnde: Forgot to update this. Will update along docstring updates.

Scripts/models/logistics.py
```diff
@@ -304,41 +339,39 @@ def run_trade_model(model: TradeRouteModule, demand: np.ndarray):
...
+        Mode (truck/container_ship...) : 2D numpy array
     """
-    n_zones = int(flow.shape[1] / len(modes))
-    flow_matrices = {}
-    for i, mode in enumerate(modes):
-        col = slice(i * n_zones, (i + 1) * n_zones)
-        flow_matrices[mode] = flow[:, col]
-    return flow_matrices
+    split_flows = np.split(flow, len(modes), axis=1)
+    mode_flows = dict(zip(modes, split_flows))
```
zptro, line 376: Now I actually understand what this does, so in that sense this is a great update. It made me think though, that separating modes as own axes in `route_demand` might streamline the whole calculation process, but that can be considered an idea for future development.

zptro (approved):

Scripts/models/logistics.py
Konnde: I too thought that the old design was too confusing and I just hadn't realized that numpy split exists. I also thought that but decided to keep the design intact for this version.

PR #364, opened 2026-02-27
Add mode name to assignment class lookup tables

attesn:
* Attempt to clarify transformations between choice model mode names and assignment classes. 
* Use lookup table in place of hard coded mode name transformations.


zptro (approved):
Nice improvement!

PR #367, opened 2026-03-17
Update transit assignment parameters

attesn:


attesn:
@jalovaj2 Ready for first test run.

Scripts/parameters/demand/ob_other.json
```diff
@@ -233,34 +233,34 @@
...
             },
             "individual_dummy": {}
         },
         "scaling": {
             "car_drv": {
-                "constant": 0.2114,
+                "constant": 0.2177,
                 "individual_dummy": {
-                    "ob_other_parent_car_drv_share": 0.2114
+                    "ob_other_parent_car_leisure_share": 0.2177
```
attesn: I forgot to replace "car_leisure" with "car_drv" in templates. This label should stay same as in `dev`. Same applies to individual dummies in all non-home-based purposes.

@jalovaj2

attesn:
This PR includes assignment parameters update and new demand parameters estimated with new impedance matrices. Aim of these changes is to get number of transfers adjusted to correct level and rail to be more attractive compared to buses, as base model underestimates transit demand in HSL region and overestimates other regions.

Overall effects of this PR are small. Transfer rate is now correct and demand in HSL region increases compared to other transport authorities (Föli, Nysse, OLS). However, this does not fix all problems of underestimated transit demand in HSL region.

@jalovaj2 @zptro

zptro (approved):

Scripts/parameters/assignment.py
zptro, line 229: I would prefer moving the "floor" penalty to the transfer penalty side. This PR seems to imply that the true transfer penalty is 10 min. Only buses would then have their own boarding penalties. However, this change would affect the mode-specific constant, so we can save it for a later update.

attesn: Yep, should have explained this change. I was thinking that having no boarding penalty would lead to trams and buses to become too attractive in short access distances. Example: Tram with 15 km/h speed and 10 min headway versus walk with 5 km/h speed. Tram becomes more attractive after 400 meters. I'd think walk would be preferred to around 800-1000 meters. 


kuva

zptro: Is not that what we have the ASC for in estimation?

attesn: I'd think access distance to train/metro is mainly result of transit assignment parameters. ASC in estimation affects the choice of walk or transit as main mode of tour, not the choice of access or egress modes in transit tour.

zptro: With this new specification a tram-metro trip chain will have 5 (tram) +5 (metro) + 5 (transfer) = 15 min total boarding penalty. With direct walk access to the metro, the penalty is 5 min. So the difference in penalty is 10 min.

If tram and metro would have no mode-specific penalty and the transfer penalty would be raised to 10 min, a tram-metro trip chain would have 10 min penalty, while direct walk access to the metro would have no penalty. Hence, also in this case the difference in penalty is 10 min.

attesn: Thanks for clarification and good point. I'll merge this and we'll floor the dummies next time we update the parameters, as estimation + model run always takes a lot of working time.

PR #368, opened 2026-03-17
Fix network preparation method parameters

Konnde:


Scripts/tests/emme_only/test_assignment.py
```diff
@@ -81,17 +82,14 @@ def __init__(self):
...
             emme_context, scenario_num, "koko_suomi",
             use_free_flow_speeds=True, time_periods={"vrk": "WholeDayPeriod"})
-        self.dist_cost = {
-            "car": 0.12,
-            "trailer_truck": 0.5,
-            "semi_trailer": 0.4,
-            "truck": 0.3,
-            "van": 0.2,
-        }
+        with open(TEST_DATA_PATH / "Scenario_input_data" / "costdata.json") as file:
```
zptro: To reduce the number of hard-coded dependencies, we could import the whole path from `tests.integration.test_data_handling`.

```suggestion
        with open(COST_DATA_PATH) as file:
```

zptro (dismissed):
Seems good to me!

Scripts/tests/emme_only/test_assignment.py
Konnde: Didn't realize test_data_handling had all of those variables, good info.

zptro (approved):

PR #370, opened 2026-03-24
Fix error caused by NoneType calibration data paths

attesn:


Scripts/travel_iteration.py
jalovaj2:
```suggestion
                 mode_dest_calibration_path: Optional[str] = None,
```

jalovaj2:
```suggestion
                 municipality_calibration_path: Optional[str] = None,
```

jalovaj2 (changes requested):
This works but we should probably change the type hints for `ModelSystem` initialization, if we plan to use string paths.

I'm not exactly sure why the `model_args` must have these paths as strings. In the future, we should probably refactor so that all paths are always `Path` objects

jalovaj2 (approved):

PR #371, opened 2026-03-24
Force trucks to pass through designated leg2 border pairs

Konnde:
Current impedance design allows trucks on second leg to incorrectly route from/to ports and between incorrect border pairs. Instead in second leg trucks are only allowed to route through specific land border control pairs.

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -774,6 +774,20 @@ class FreightPurpose(Purpose):
             self.costdata, ship_imps, self.model_category, fin_border_ids,
             self.is_export)
         impedance_legs["leg_two"].update(ship_costs)
+
+        # Retain leg two truck cost only for designated border pairs
+        mask = numpy.ones(impedance_legs["leg_two"]["truck"]["cost"].shape, dtype=bool)
+        fin_indices = {key: idx for idx, key in enumerate(fin_border_ids)}
+        foreign_indices = {key: idx for idx, key in enumerate(cluster_border_ids)}
+        for border_pair in param.land_border_pairs.values():
+            if (border_pair["finland_border"] in fin_border_ids 
+            and border_pair["foreign_border"] in cluster_border_ids):
+                fin_index = fin_indices[border_pair["finland_border"]]
+                foreign_index = foreign_indices[border_pair["foreign_border"]]
+                row = fin_index if self.is_export else foreign_index
+                col = foreign_index if self.is_export else fin_index
+                mask[row, col] = False
+        impedance_legs["leg_two"]["truck"]["cost"][mask] = numpy.inf
         return impedance_legs
 
     def get_costs(self, impedance: dict):
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -775,17 +775,14 @@ class FreightPurpose(Purpose):
             self.is_export)
         impedance_legs["leg_two"].update(ship_costs)
 
-        # Retain leg two truck cost only for designated border pairs
-        pairs = numpy.array(list(param.land_border_pairs.values()))
-        exists = pairs[numpy.isin(pairs[:, 0], fin_port_zones) 
-                       & numpy.isin(pairs[:, 1], cluster_port_zones)]
-        fin_idx = numpy.searchsorted(fin_port_zones, exists[:, 0])
-        cluster_idx = numpy.searchsorted(cluster_port_zones, exists[:, 1])
-        mask = numpy.ones(impedance_legs["leg_two"]["truck"]["cost"].shape, dtype=bool)
-        rows = fin_idx if self.is_export else cluster_idx
-        cols = cluster_idx if self.is_export else fin_idx
-        mask[rows, cols] = False
-        impedance_legs["leg_two"]["truck"]["cost"][mask] = numpy.inf
+        # Retain leg two truck cost only for designated land border pairs
+        mask = pandas.DataFrame(True, index=fin_port_zones, columns=cluster_port_zones)
+        for fin_border, cluster_border in param.land_border_pairs.values():
+            if fin_border in fin_port_zones and cluster_border in cluster_port_zones:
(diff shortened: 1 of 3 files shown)


Scripts/datatypes/purpose.py
```diff
@@ -774,6 +774,20 @@ def form_impedance_legs(self, impedance: dict,
             self.costdata, ship_imps, self.model_category, fin_border_ids,
             self.is_export)
         impedance_legs["leg_two"].update(ship_costs)
+
+        # Retain leg two truck cost only for designated border pairs
+        mask = numpy.ones(impedance_legs["leg_two"]["truck"]["cost"].shape, dtype=bool)
+        fin_indices = {key: idx for idx, key in enumerate(fin_border_ids)}
+        foreign_indices = {key: idx for idx, key in enumerate(cluster_border_ids)}
```
zptro: You could consider using a pandas DataFrame with label-based indexing as `mask` instead of this multi-step indexing manoevre.

Scripts/datatypes/purpose.py
```diff
@@ -774,6 +774,18 @@ def form_impedance_legs(self, impedance: dict,
...
+        pairs = numpy.array(list(param.land_border_pairs.values()))
+        exists = pairs[numpy.isin(pairs[:, 0], fin_port_zones) 
+                       & numpy.isin(pairs[:, 1], cluster_port_zones)]
+        fin_idx = numpy.searchsorted(fin_port_zones, exists[:, 0])
+        cluster_idx = numpy.searchsorted(cluster_port_zones, exists[:, 1])
+        mask = numpy.ones(impedance_legs["leg_two"]["truck"]["cost"].shape, dtype=bool)
+        rows = fin_idx if self.is_export else cluster_idx
+        cols = cluster_idx if self.is_export else fin_idx
+        mask[rows, cols] = False
+        impedance_legs["leg_two"]["truck"]["cost"][mask] = numpy.inf
```
Konnde: I also thought of dataframe alternative but it's pretty much the same

```
        fin, cluster = "fin_border", "cluster_border"
        df = pandas.DataFrame(impedance_legs["leg_two"]["truck"]["cost"],
                              index=fin_port_zones, columns=cluster_port_zones)
        pairs = pandas.DataFrame(param.land_border_pairs.values(),
                                 columns=[fin, cluster])
        exists = pairs[pairs[fin].isin(df.index) & pairs[cluster].isin(df.columns)]
        rows = exists[fin] if self.is_export else exists[cluster]
        cols = exists[cluster] if self.is_export else exists[fin]
        mask = pandas.DataFrame(True, index=df.index, columns=df.columns, dtype=bool)
        mask.loc[rows.values, cols.values] = False
        df = df.mask(mask, numpy.inf)
        impedance_legs["leg_two"]["truck"]["cost"] = df.to_numpy()
```

zptro: Great that you found a way to remove the for-loop. However, the point of using pandas would be to utilize its label-based indexing. I have not been able to test whether this works though.

```suggestion
        pairs = list(param.land_border_pairs.values())
        fin_land_zones = [pair[0] for pair in pairs if pair[0] in fin_port_zones]
        cluster_land_zones = [pair[1] for pair in pairs if pair[1] in cluster_port_zones]
        mask = pandas.DataFrame(True, index=fin_port_zones, columns=cluster_port_zones)
        mask.loc[fin_land_zones, cluster_land_zones] = False
        if not self.is_export:
            mask = mask.T
        impedance_legs["leg_two"]["truck"]["cost"][mask.to_numpy()] = numpy.inf
```

Konnde: I did a quick test with this design suggestion in municipality network along with the different clusters and borders. It seems that this design doesn't enforce pairs correctly. It still allows trips from each land border to all land border whereas each should only have one correct pair. The result looks like this

	Commodity	Type	Mode	Finnish border	Foreign border	Tons (t/annual)
0	elektr	export	truck	FIVAI	NOKIL	0.00459
1	elektr	export	truck	FIVAI	RUVAR	0.10571
2	elektr	export	truck	FIVAI	RUVAL	20.59697
3	elektr	export	truck	FIVAI	SEHAA	0.04571
...

FIVAI shouldn't actually be outputted at all because its pair node doesn't currently exist in our large network. I have informed about this. But if it did, FIVAI should only pair to RUVAI.

The summary that is outputted with the current design is correct and looks like this

	Commodity	Type	Mode	Finnish border	Foreign border	Tons (t/annual)
0	elektr	export	truck	FIVAL	RUVAL	37.70219
1	elektr	export	truck	FIVAR	RUVAR	0.78176
2	elektr	export	truck	FITOR	SEHAA	3494.42847
3	elektr	export	truck	FIKIL	NOKIL	208.95532
4	elektr	export	container_ship	FIHEL	CIABJ	0.03325

I guess what is missing from the pandas design either some missing reference or cross merge between indeces and columns to get all possible pairs.

zptro: I see. It seems that pandas array indexing works differently than numpy array indexing. I guess it is better to use numpy in this case. Thanks for testing though!

zptro: Your solution in elegant in the sense that it does not require any looping, even if the 3-step indexing is a bit difficult to understand. The only small change that I think is necessary is the name of `exists`, as the current name suggests it is a boolean array. Maybe `existing_pairs`?

However, if one for loop is acceptable, I think this would be the most elegant (and easy to understand) way of utilizing pandas.

```suggestion
        mask = pandas.DataFrame(True, index=fin_port_zones, columns=cluster_port_zones)
        for fin_border, cluster_border in param.land_border_pairs.values():
            try:
                mask.at[fin_border, cluster_border] = False
            except KeyError:
                pass
        if not self.is_export:
            mask = mask.T
        impedance_legs["leg_two"]["truck"]["cost"][mask.to_numpy()] = numpy.inf
```

Konnde: This is good. Let's replace the try-except block with if conditioning to check if pair is found on network.

zptro (approved):

PR #372, opened 2026-03-26
calc trade domestic leg share

Konnde:
In our current design structure, the domestic leg for each of our traded commodities has just one estimated mode that is truck. This is because of lack in freight train related estimation data. However we would also like to understand what is the potential that could be transported with freight train between zones and border control points. Therefore we can use modelled results of our domestic purposes to gain some understanding of distribution between truck and freight train.

Note that all of our domestic models do not have freight train as a mode alternative meaning that freight train will not become a land mode alternative for that purpose's foreign models.

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -846,23 +846,24 @@ class FreightPurpose(Purpose):
                      where=demand_sum > 0.0)
         train_share = numpy.ones_like(truck_share) - truck_share
         
-        purpose_trade_demand = {k: v for k, v in trade_demand.items()
-                                if k.split('_')[0] == self.name}
+        purpose_trade_demand = {
+            k: v for k, v in trade_demand.items()
+            if k.startswith(f"{self.name}_")
+        }
         fin_borders = numpy.array(list(fin_borders.values()), dtype=numpy.int32)
         mask = numpy.isin(self.orig_zone_numbers, fin_borders)
-        dom_leg_demand = {
-            purpose: {mode: numpy.zeros_like(truck_share)
-                      for mode in ["truck", "freight_train"]} 
-            for purpose in purpose_trade_demand
-        }
+        
+        dom_leg_demand = {}
         for purpose in purpose_trade_demand:
             demand_full = numpy.zeros_like(truck_share)
-            if purpose.split("_")[1] == "export":
+            if purpose.endswith("_export"):
                 demand_full[:, mask] = purpose_trade_demand[purpose]["leg_one"]["truck"]
             else:
                 demand_full[mask, :] = purpose_trade_demand[purpose]["leg_three"]["truck"]
-            dom_leg_demand[purpose]["truck"] += demand_full * truck_share
-            dom_leg_demand[purpose]["freight_train"] += demand_full * train_share
+            dom_leg_demand[purpose] = {"truck": demand_full * truck_share}
+            train_share = demand_full * train_share
+            if numpy.sum(train_share) > 0.0:
+                dom_leg_demand[purpose].update({"freight_train": train_share})
         return dom_leg_demand
 
     def run_logistics_module(self, demand_truck: numpy.ndarray, impedance: numpy.ndarray, 
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -826,7 +826,7 @@ class FreightPurpose(Purpose):
         demand : dict
             Mode (truck/train/...) : calculated demand (numpy 2d matrix)
         trade_demand : dict
-            Purpose name (elektr_export/elektr_import...) : Leg
+            Foreign FreightPurpose : Leg
                 Name (one/two/three) : Mode
                     Name (truck/container_ship...) : numpy 2d array
         fin_borders : dict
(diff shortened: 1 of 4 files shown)


Scripts/datatypes/purpose.py
```diff
@@ -816,6 +816,53 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
 
+    def calc_trade_mode_share(self, demand: dict, trade_demand: dict,
+                              fin_borders: dict):
+        """Calculate mode share of trade demand on its domestic leg 
+        between freight land modes.
+
+        Parameters
+        ----------
+        demand : dict
+            Mode (truck/train/...) : calculated demand (numpy 2d matrix)
```
zptro: I think it would be good to clarify that this is `domestic_demand`

Scripts/datatypes/purpose.py
```diff
@@ -816,6 +816,53 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
+        trade_demand : dict
+            Foreign FreightPurpose : Leg
+                Name (one/two/three) : Mode
```
zptro: This seems strange. What is a `Leg` and a `Mode`?

Scripts/datatypes/purpose.py
```diff
@@ -816,6 +816,53 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
+        """
+        demand_sum = sum(demand[mode] for mode in demand
+                         if mode in ["truck", "freight_train"])
+        truck_share = numpy.zeros_like(demand_sum)
+        numpy.divide(demand["truck"], demand_sum, out=truck_share,
+                     where=demand_sum > 0.0)
+        train_share = numpy.ones_like(truck_share) - truck_share
+        
+        purpose_trade_demand = {k: v for k, v in trade_demand.items()
+                                if k.name.split('_')[0] == self.name}
```
zptro: So you are now using the `Purpose` instance itself as key? How can that even work? By the way, where is the unit test?

zptro (changes requested):

Scripts/datatypes/purpose.py
Konnde: Updated docstrings

Scripts/datatypes/purpose.py
Konnde: Switched key from FreightPurpose object to foreign purpose.name (str). Expanded FreightModelTest unittest to consider this method.

Scripts/datatypes/purpose.py
```diff
@@ -816,6 +816,56 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
+                         if mode in ["truck", "freight_train"])
+        truck_share = numpy.zeros_like(demand_sum)
+        numpy.divide(dom_demand["truck"], demand_sum, out=truck_share,
+                     where=demand_sum > 0.0)
+        train_share = numpy.ones_like(truck_share) - truck_share
+        
+        purpose_trade_demand = {
+            k: v for k, v in trade_demand.items()
+            if k.startswith(f"{self.name}_")
+        }
```
zptro, line 853: Could you add a comment with an example of what trade purposes this gathers?

Scripts/datatypes/purpose.py
```diff
@@ -816,6 +816,56 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
+        ----------
+        dom_demand : dict
+            Mode (truck/train/...) : domestic demand numpy 2d array
+        trade_demand : dict
+            Foreign purpose name : str
+                Leg (one/two/three) : str
+                    Mode (truck/container_ship...) : trade demand numpy 2d array
+        fin_borders : dict
+            Finland border id (FIHEL/FISKV...) : str
+                Centroid id : int
```
zptro: What you really want is only a list of centroid IDs. Would it be more clear then to extract that list outside this function?

Scripts/datatypes/purpose.py
```diff
@@ -816,6 +816,56 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
+        for purpose in purpose_trade_demand:
+            demand_full = numpy.zeros_like(truck_share)
+            if purpose.endswith("_export"):
+                demand_full[:, mask] = purpose_trade_demand[purpose]["leg_one"]["truck"]
+            else:
+                demand_full[mask, :] = purpose_trade_demand[purpose]["leg_three"]["truck"]
+            dom_leg_demand[purpose] = {"truck": demand_full * truck_share}
+            train_share = demand_full * train_share
+            if numpy.sum(train_share) > 0.0:
+                dom_leg_demand[purpose].update({"freight_train": train_share})
```
zptro:
```suggestion
                dom_leg_demand[purpose]["freight_train"] = train_share
```

Scripts/datatypes/purpose.py
```diff
@@ -816,6 +816,56 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
+        
+        dom_leg_demand = {}
+        for purpose in purpose_trade_demand:
+            demand_full = numpy.zeros_like(truck_share)
+            if purpose.endswith("_export"):
+                demand_full[:, mask] = purpose_trade_demand[purpose]["leg_one"]["truck"]
+            else:
+                demand_full[mask, :] = purpose_trade_demand[purpose]["leg_three"]["truck"]
+            dom_leg_demand[purpose] = {"truck": demand_full * truck_share}
+            train_share = demand_full * train_share
```
zptro: You are using `train_share` both as a constant multiplier and as the dynamically changing result of the multiplication. I guess that is not what is intended?

Scripts/datatypes/purpose.py
```diff
@@ -816,6 +816,56 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
+        fin_borders = numpy.array(list(fin_borders.values()), dtype=numpy.int32)
+        mask = numpy.isin(self.orig_zone_numbers, fin_borders)
+        
+        dom_leg_demand = {}
+        for purpose in purpose_trade_demand:
+            demand_full = numpy.zeros_like(truck_share)
+            if purpose.endswith("_export"):
+                demand_full[:, mask] = purpose_trade_demand[purpose]["leg_one"]["truck"]
+            else:
+                demand_full[mask, :] = purpose_trade_demand[purpose]["leg_three"]["truck"]
```
zptro: If `demand_full` would be a `DataFrame`, you would not need to construct a separate `bool` mask, because you could use `fin_borders` directly in `loc`. But this is more of a taste thing.

zptro (changes requested):
This looks very good, just some small remarks. The summary utils could maybe be revised a bit at some point, but that is not very critical.

Konnde:
Addressed noted comments. Added summary writing functions into test_freight_demand unittest for users to see txt-file outputs when using test environment.

Scripts/utils/freight_utils.py
```diff
@@ -86,26 +86,22 @@ def write_domestic_leg_summary(demand_trade: dict, impedance: dict,
...
+            "Type": trade_type,
+            "Mode": mode,
+            "Tons (t/annual)": numpy.sum(demand_trade[purpose][mode], dtype=numpy.float32),
+            "Ton mileage (tkm/annual)": numpy.sum(
+                demand_trade[purpose][mode] * impedance[mode]["dist"], dtype=numpy.int64),
+        }
+        for purpose in demand_trade
+        for name, trade_type in [purpose.split("_")]
+        for mode in demand_trade[purpose]
+    ]
```
zptro, line 102: Nice!

zptro (approved):

PR #373, opened 2026-04-09
Create gen_cost matrix only when needed, and pop time and cost

zptro:
The point is to save memory.

N.B. I was considering a more elegant solution with the gen_cost matrix creation on the assignment side (where it actually already exists), but that would have made the purpose-wise cost sharing calculations impossible.


attesn (approved):
Good point!

PR #374, opened 2026-04-10
Car access allowed only directly to long-distance transit modes

zptro:
This seems to speed up long-distance assignment by ~1h (~6h ->~5h), hence ~10min per car-access class. 

This PR affects route choices in Helsinki region somewhat, because before this new restriction it has made sense to minimize parking costs by transferring via commuter train. The effect on total impedances should not be large and the new route choices are overall probably more realistic.


attesn:
@zptro Added new estimated parameters. Overnight tours model performs better than before. Business tours model is not valid in either case, but that will be solved in future work.

attesn (dismissed):

attesn (approved):

PR #376, opened 2026-04-15
Implement foreign passenger models

jalovaj2:
Implements 2 separate foreign passenger models to the modelsystem:

-------------------------------------------------------------------------------

1. Foreign external passenger model
- Models external passenger traffic to "external" purpose zones which are not included in the zonedata when modelling long distance trips.

The proposed workflow:
1. The model uses an external demand matrix (full matrix) imported as a separate .omx file from base matrices. The matrix should contain nonzero entries only between "domestic" and "external" zones.
2. The generation from the demand matrix is calibrated according to ("domestic") origin zone variables (for example, 'population). Attraction is not calibrated since we don't have zonedata for the ("external") destinations. 
3. The demand matrix is balanced to match the calibrated generation using Fratar matrix balancing.
4. The demand matrix is then split to different (airplane) access modes based on the access mode probabilities calculated from the access mode choice models. To do this, we use Zonedata object that contains zone data entries for each zone in the model such that the zones without zone data get default values (0 or inf depending on the context) for each entry.
5. The resulting access mode-specific demand matrices are added to the model and assigned normally together with other external demand during the base demand assignment.

-----------------------------------------------------------------------------

2. Foreign northern passenger model
- Models external passenger traffic to "foreign" purpose zones when modelling long distance trips.

The proposed workflow:
1. We create a new travel purpose named hb_abroad_northern. This travel purpose works exactly the same way as other long distance travel groups with the exception that the generation area is "domestic" and attraction area is "foreign". Currently, the parameters for this travel purpose are copied from hb_leisure_overnight.

Note: Using different purpose areas ("domestic and "foreign") for generation and attraction in logit models seems not to be supported yet. Significant modifications have been made to both logit.py and zonedata.py to accept different purpose areas and their related zonedata in the logit formulations.

------------------------------------------------------------------------------


jalovaj2:
This Pull Request is still a work in progress, since there are multiple errors in the implementation waiting to be fixed. Here's a list of the most apparent ones:

1. When calculating mode-specific demand matrices for Foreign northern passenger model (hb_abroad_northern.json), the demand_vrk.omx entries associated with "foreign" zones get NaN values. This is probably due to missing zonedata entries that result in utilities with -inf or inf values. These have to be revised to make sure the utilities stay reasonable even with default values.

2. When trying to add mode-specific demand matrices from Foreign external passenger model, seemingly no demand is actually added to the model, even though the values at the matrices seem reasonable. This is probably a bug related to the zone slicing when adding external demand.

3. When testing the short distance models, The car_drv share slightly changes from the test validation share. This indicates that the logic for separate generation and attraction in logit models is faulty and doesn't reproduce the same results if the generation and attraction areas are both "domestic". The logit models have to be revised for the new logic that includes different generation and attraction zones.

jalovaj2:
I will split this pull request to to 2 different pull requests, one Foreign external passenger model and one for Foreign northern passenger model

PR #377, opened 2026-04-15
Remove unnecessary LOS matrices and model specification syntax

zptro:
- Remove "transform" syntax from long-distance model specifications. The syntax is used when variables have to be summed up before logarithmized, but a single variable can be logarithmized directly.
- Remove `car_time` and `transfer_time` handling from demand model, as they no longer have purpose-specific calculations or parameters. The `car_time` weight correction is more clear and simple to do on the assignment side.


attesn (dismissed):
Much better syntax!

zptro:
If we merge this now, it will be in effect in upcoming @jonisylvinwsp LOS matrices. This means that from that estimation onward, LOS transformations need to be adjusted on estimation side, @attesn.

attesn (approved):

PR #378, opened 2026-04-16
Fix directions for access and egress modes

zptro:
When separate car-egress modes were added in #308, the access-egress departure time logic was not changed accordingly. In #350 car-egress modes were added, but not with the trips going in the right direction. This should fix the problem.


Scripts/assignment/departure_time.py
```diff
@@ -95,14 +95,14 @@ def add_demand(self, demand: Demand):
         """
         position: Sequence[int] = demand.position
         ass_classes = mode_assignment_classes[demand.mode]
-        for ass_class in ass_classes:
+        for is_return, ass_class in enumerate(ass_classes):
```
attesn, line 98: I’m a bit worried about this condition. If I'm correct, the logic depends on the order of ass_classes in parameters, which is easy to miss, as there is also the integer as boolean treatment in same condition.

It would be good to describe the requirement of ass_class order in parameters or make the condition more explicit in the code, altough it probably wont be as clean as this solution.

attesn (dismissed):
Good that you spotted and fixed this. One comment below.

Scripts/assignment/departure_time.py
zptro: Good point! I did not come up with a new solution, but added a clarifying comment.

attesn (approved):

PR #381, opened 2026-04-17
Implement foreign external passenger model

jalovaj2:
Implements foreign external passenger model to the modelsystem. The model is used to model external passenger traffic to "external" purpose zones which are not included in the zonedata when modelling long distance trips.

The proposed workflow:

1. The model uses an external demand matrix (full matrix) imported as a separate .omx file from base matrices. The matrix should contain nonzero entries only between "domestic" and "external" zones.
2. The generation from the demand matrix is calibrated according to ("domestic") origin zone variables (for example, 'population'). Attraction is not calibrated since we don't have zonedata for the ("external") destinations.
3. The demand matrix is balanced to match the calibrated generation using Fratar matrix balancing.
4. The demand matrix is then split to different (airplane) access modes based on the access mode probabilities calculated from the access mode choice models. To do this, we use Zonedata object that contains zone data entries for each zone in the model such that the zones without zone data get default values (0 or inf depending on the context) for each entry.
5. The resulting access mode-specific demand matrices are added to the model and assigned normally together with other external demand during the assignment.


Scripts/datatypes/purpose.py
zptro, line 429: One idea would be to refactor this method so that we could skip this step entirely in the foreign external model.

Scripts/demand/foreign_external.py
```diff
@@ -0,0 +1,81 @@
...
+            mapping_full = mtx.mapping
+            # Remove zero values
+            base_mtx = mtx[mode].clip(0.000001, None)
+            base_colsum = base_mtx.sum(1)
+            # Add ones for missing zones in zonedata
+            production_base_full = numpy.array([0 if i not in self.zone_numbers_zone_data else production_base[i] for i in self.zone_numbers_full])
+            production_forecast_full = numpy.array([0 if i not in self.zone_numbers_zone_data else production_forecast[i] for i in self.zone_numbers_full])
+        production = calibrate(
+            base_colsum, production_base_full, production_forecast_full)
+        mtx = pandas.DataFrame(base_mtx, self.zone_numbers_full, self.zone_numbers_full)
```
zptro: Here I think it would be more simple and straightforward to import and use an n x m matrix instead of n+m x n+m.

Scripts/travel_iteration.py
```diff
@@ -321,6 +337,30 @@ def assign_base_demand(self,
...
+                # Access mode demand
+                for access_mode in assignment_mode_probs:
+                    access_probs = assignment_mode_probs[access_mode]
+                    access_mode_mtx = foreign_ext_mtx * access_probs
+                    access_mode_demand = Demand(purp_abroad_other, access_mode, access_mode_mtx)
+                    self.dtm.add_demand(access_mode_demand)
+                    mtx[access_mode] = access_mode_demand.matrix
+                # Main mode demand (airplane)
+                mtx["airplane"] = foreign_ext_mtx
+            log.info("Foreign external demand added and saved to " + str(self.resultmatrices.path))
```
zptro: I suggest moving all of this to `calc_prob` in a sub-class of `TourPurpose`. In that way it could be run in the same loop as the other purposes. And I do not think it is necessary to store the resulting matrix in a separate file, because it will be clearly separable from other long trips based on origin-destination pairs.

Scripts/travel_iteration.py
```diff
@@ -300,9 +315,10 @@ def assign_base_demand(self,
                 self.freight_matrices, param.truck_classes)
 
         # Add beeline distance dummy
-        zd = self._zone_datas["domestic"]
-        idx = numpy.isin(self.zone_numbers, zd.zone_numbers)
-        zd["beeline"] = Purpose.distance[numpy.ix_(idx, idx)]
+        for model_area in self._zone_datas:
+            zd = self._zone_datas[model_area]
+            idx = numpy.isin(self.zone_numbers, zd.zone_numbers)
+            zd["beeline"] = Purpose.distance[numpy.ix_(idx, idx)]
```
zptro: I do not think beeline distance is a relevant variable in the external demand model.

Scripts/travel_iteration.py
```diff
@@ -106,6 +108,15 @@ def __init__(self,
...
             ) for model_area in ["domestic"]}
+        
+        # Create zone data for all zones, created from domestic zonedata and default values for zones not in .gpkg
+        zonedata_all = ZoneData(
+            zone_data_path, self.zone_numbers, submodel,
+            model_area="domestic", municipality_calibration=municip_calib,
+            extra_dummies=extra_dummies,
+            car_dist_cost=self.car_dist_cost["car"])
+        zonedata_all.reindex_zones(self.zone_numbers)
+        self._zone_datas["all"] = zonedata_all
```
zptro: Not necessary, because we do not have or need zone data for external areas.

Scripts/utils/external.py
```diff
@@ -0,0 +1,63 @@
...
+        Calibrated forecast
+    """
+    b = calib_base
+    n = production_base
+    s = production_forecast
+    threshold = param.vector_calibration_threshold
+    n[n == 0] = 0.000001
+
+    debug = numpy.where(s < threshold*n, s * b/n, s + threshold*(b - n))
+
```
zptro, line 60: Remember to remove.

Scripts/parameters/departure_time.py
```diff
@@ -65,6 +65,18 @@
...
+            "aht": (0.066, 0),
+            "pt": (0.07, 0),
+            "iht": (0.066, 0),
+        },
+        "airplane": {
+            "aht": (0.066, 0),
+            "pt": (0.07, 0),
+            "iht": (0.066, 0),
+        },
+    },
```
zptro: I do not think we need separate demand shares for the foreign external demand. In the hourly traffic assignments, it will be part of long-distance demand matrix (external).

Scripts/parameters/ext_tour_generation.py
zptro, line 1: You have the json file for these things, so why create a new file?

zptro (commented):
Some suggestions on how to restructure the code below. The main point is to not use `ZoneData` for the external zones, but instead specify simple weighting factors for them.

PR #382, opened 2026-04-17
Implement foreign northern passenger model

jalovaj2:
Implements foreign northern passenger model to the modelsystem. The model is used to model passenger traffic to "foreign" purpose zones when modelling long distance trips.

The proposed workflow:

1. We create a new travel purpose named hb_abroad_northern. This travel purpose works exactly the same way as other long distance travel groups with the exception that the generation area is "domestic" and attraction area is "foreign". Currently, the parameters for this travel purpose are copied from hb_leisure_overnight.


jalovaj2:
Note: The foreign entries in the demand matrix currently have NaN values. This is probably due to missing zonedata for foreign zones, resulting in abnormal utilities in the logit model, rendering NaN values to the matrix. The next step is to figure out the functional default values for foreign zonedata and, if necessary, the required zonedata for foreign zonedata.

PR #384, opened 2026-04-23
Update import coefficients for some commodities

IlpoRatinen:
- Updated coefficients and dummies for imported commodity groups minkai, metkai, kokbri and krhiil in foreing freight model


Konnde (changes requested):
Tested these updates with municipality level model and did not encounter problems. Two extra indents on krhiil_import and metkai_import. Also title indicates export but changed files are import.

Konnde (approved):
Looks good.

PR #385, opened 2026-04-23
Reuse cost calculation from truck route choice in later stage

zptro:
This is the first step in streamlining freight cost calculations. Instead on re-calculating distance-based and time-based truck costs before demand calculations, the vehicle-level cost matrix is taken directly from assignment (with its route choice distance and time cost assumptions) and then transformed to goods-ton level with "avg_load" factor.

Also refactored tests a bit.


Scripts/assignment/datatypes/freight_specification.py
```diff
@@ -118,10 +140,12 @@ def get_matrices(self):
             **self.dist.item,
             **self.aux_dist.item,
             **self.time.item,
-            **self.aux_time.item,
             **self.canal_cost.item,
         }
+        aux_cost = (self._time_unit_cost*self.time.data/60
+                    + self._dist_unit_cost*self.dist.data)
```
zptro: Ah, should be `self.aux_time` and `self.aux_dist`. Will fix this.

Scripts/utils/freight_costs.py
```diff
@@ -62,8 +63,11 @@ def calc_road_cost(unit_costs: Dict[str, Dict],
     """
     mode_cost = {}
     for mode, params in unit_costs["truck"].items():
-        mode_cost[mode] = ((sum(impedance[k] * params[k] for k in impedance) 
-                           + params["terminal_cost"])
+        mode_cost[mode] = (((1+params["empty_share"])
+                            * truck_overhead_cost
+                            * impedance["cost"]
```
Konnde, line 68: Raises KeyError here for costs here while running through valma freight. I think this is linked to cost and aux_cost not being defined here
https://github.com/Traficom/valma-model-system/blob/959fcc9fa9c9c81dd0686d09678b304d4cfcea78/Scripts/assignment/freight_assignment.py#L47-L48

Konnde (changes requested):
Been very busy during this latest week so I haven't been able to review before now. I did glance over the changes and they seem ok. I did find one bug that needs to be fixed.

The last thing I am going to review first thing next week are cost outputs of this branch versus cost outputs of dev. Will inform more then.

Scripts/utils/freight_costs.py
zptro: Thanks for checking this! Now this should be fixed.

Konnde (approved):
After some more testing where costs were cost differences were still present, we looked into this with @zptro. We figured that the unit costs in the updated cost json influence route choice causing change in some od pair impedances. As this is a desired effect, we can go ahead with merging.

PR #388, opened 2026-04-24
Refactor handling of matrices and dimensions in `ZoneData`

zptro:
- @jalovaj2 discovered that generation and attraction zone data were not handled properly in logit models. This PR fixes that. At the same time, the `bounds` and `generation` arguments in `ZoneData.get_data()` have become obsolete, as we now handle that with the separate generation and attraction `ZoneData` instances. So I removed `ZoneData.get_data()` altogether and replaced it with `ZoneData.__get_item__()`
- Make 2-d matrix handling in `ZoneData` more memory efficient and straightforward.


jalovaj2 (approved):
Looks good!

PR #389, opened 2026-04-24
Expand OD-validation of LOS matrices

attesn:
Write validation data of LOS matrices to file, which can be opened in Excel-format and compared to previous model runs to find differences and outliers. We can guide modellers to use this functionality to validate their assignment runs.

Speed is still logged to terminal, as zero or inf would suggest that model run should be interrupted.


Scripts/travel_iteration.py
```diff
@@ -314,6 +315,18 @@ def assign_base_demand(self,
...
+                for ass_class in impedance[tp][mtx_type]:
+                    output_od_los(impedance[tp][mtx_type][ass_class], self.zone_numbers, 
+                                  mtx_type, ass_class, tp, self.resultdata)
+            for ass_class in self.ass_classes:
+                try:
+                    dist = impedance[tp]["dist"][ass_class]
+                    time = impedance[tp]["time"][ass_class]/60
+                    log_avg_speed(dist, time, ass_class, tp)
+                except KeyError:
+                    pass
```
zptro: I suggest hiding this in a function in `validate_assignment` module. Both `output_od_los` and `log_avg_speed` are also so short that maybe they both could be part of that function?

Scripts/utils/validate_assignment.py
```diff
@@ -1,41 +1,61 @@
...
     """
-    los = int(los_mtx[mapping[202], mapping[17278]])
-    los = f"{los//60}h {los%60}min" if "time" in mtx_type else los
-    log.debug(f"{mtx_type} {mode} Helsinki - Jyvaskyla: {los}")
+    zone_to_idx = {zone: idx for idx, zone in enumerate(zone_numbers)}
+    for od_pair, (z0, z1) in print_od_pairs.items():
+        idx0 = zone_to_idx.get(z0)
+        idx1 = zone_to_idx.get(z1)
+        if idx0 is None or idx1 is None:
+            continue
```
zptro:
```suggestion
        try:
            idx0 = zone_to_idx[z0]
            idx1 = zone_to_idx[z1]
        except KeyError:
            continue
```

zptro (changes requested):
Very good refactoring and generalization! I suggest taking it one step further:

attesn:
We could move this to post-demand-model-run scripts in near future. Then it is easier to adjust printed OD-pairs.

zptro (approved):
Very good improvement!

PR #392, opened 2026-04-30
Remove utility matrices used in old agent model

zptro:
This will free some more memory.


jonisylvinwsp (approved):
This seems promising, good to have for next runs especially incase the current run would fail at the end.

PR #393, opened 2026-05-04
fix matrix position when adding asymmetric demand

jalovaj2:
Fixes a crash that occurs when demand is added to the DepartureTimeModel from egress assignment classes for purposes with different generation and attraction zones.

### DIFF
--- a/Scripts/assignment/departure_time.py
+++ b/Scripts/assignment/departure_time.py
@@ -101,7 +101,7 @@ class DepartureTimeModel:
                 for ap in self.assignment_periods:
                     if ass_class in ap.assignment_modes:
                         mtx = demand.matrix.T if is_return else demand.matrix
-                        position = (demand.position[1], demand.position[0]) if is_return else demand.position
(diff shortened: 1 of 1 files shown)


Scripts/assignment/departure_time.py
```diff
@@ -101,6 +101,7 @@ def add_demand(self, demand: Demand):
                 for ap in self.assignment_periods:
                     if ass_class in ap.assignment_modes:
                         mtx = demand.matrix.T if is_return else demand.matrix
+                        position = (demand.position[1], demand.position[0]) if is_return else demand.position
```
zptro:
```suggestion
                        position = demand.position[::-1] if is_return else demand.position
```

Scripts/assignment/departure_time.py
zptro, line 96:
```suggestion
```

zptro (changes requested):
Good catch! You could perhaps make the syntax a bit more pythonic?

zptro (approved):

PR #394, opened 2026-05-04
Streamline tour generation, especially non-home

zptro:
The `NonHomeGeneration` model also creates `f"{self.purpose.name}_parent_{mode}_share"` zone variables needed in mode choice model. Previously they were not available in first probability calculation attempt, resulting in the complicated stashing logic in `models.logit` module. Now when generation is independent for each tour purpose, we can always do the zone variable creation before mode choice, resulting in less dead weight.

For home-based generation there is no need to calculate tour numbers in advance.


attesn (dismissed):
Nice!

attesn (approved):

PR #395, opened 2026-05-05
Fix park-and-ride cost

zptro:
This fixes handling of parking cost in park-and-ride modes both on assignment and demand side.

Assignment side: Add parking cost to cost matrix. This means that time matrix, which is calculated by removing cost from generalized cost, changes as well.

Demand side: Add parking cost to generalized cost where it should be.


attesn (approved):
Good fix, I'll need to see how distance comparisons and mode share validation changes.

attesn (approved):

PR #396, opened 2026-05-05
Add secondary destination volumes to assignment matrices

attesn:
Use rate calculated from added beeline distance from detour


zptro (approved):
Nice and clean!

PR #398, opened 2026-05-06
Add vans based on car volumes only

attesn:
* Add vans based on car volumes only as truck traffic development is ongoing.
* Update rates based on Statistics Finland 2016 summary statistics on van and car mileage. 
* Assumed that vans have even time distribution and car peaks follow HLT 2016 data.
* Calculation of new rates is stored in Sharepoint `05_Henkilöliikenne / Pakettiautot`.


zptro (approved):
The pt share might be a bit exaggerated compared to aht, but overall this seems valid!

PR #399, opened 2026-05-07
Update demand parameters with new car costs

attesn:
Demand parameters estimated with car km-cost that accounts for car fleet composition and car depreciation.

Things to do when merging:
* Replace `--dev` cost and omx-files in Sharefile with new ones (`--cost`). 
* Send mail to active model users that they need to update their cost-data if using `dev` branch.


zptro (approved):
Super fast!

PR #401, opened 2026-05-07
Implement foreign external passenger model

jalovaj2:
Implements foreign external passenger model to the modelsystem. The model is used to model external passenger traffic to "external" purpose zones which are not included in the zonedata when modelling long distance trips.

The proposed workflow:

1. The model uses an external demand matrix imported as a separate .omx file from base matrices. The matrix contains entries between "domestic" and "external" zones.
2. The generation from the demand matrix is calibrated according to ("domestic") origin zone variables (for example, 'population'). Attraction is not calibrated since we don't have zonedata for the ("external") destinations.
3. The demand matrix is then split to different access modes based on the access mode probabilities calculated from the access mode choice models.
4. The resulting access mode-specific demand matrices are added to the model and assigned normally together with other external demand during the assignment.


Scripts/datatypes/purpose.py
```diff
@@ -949,3 +955,109 @@ def run_trade_route_module(self, impedance: dict,
...
+        return acc_splits
+
+    def split_connection_mode(self, impedance, pt_mode):
+        if pt_mode == "airplane":
+            impedance["airpl_taxi_acc"] = impedance["airpl_car_acc"]
+        model = self.connection_models[pt_mode]
+        prob, logsum = model.calc_mode_prob(impedance)
+        if "airpl_taxi_acc" in prob:
+            prob["airpl_car_acc"] += prob.pop("airpl_taxi_acc")
+        return prob, logsum
```
zptro: This is inherited from `TourPurpose` so no need to have an own method here.

Scripts/datatypes/purpose.py
```diff
@@ -949,3 +955,109 @@ def run_trade_route_module(self, impedance: dict,
...
+                    f"logsum_{self.name}", "vrk", list(self.orig_zone_numbers), m='w'
+                    ) as mtx:
+                for main_mode, acc_modes in self.intermodals.items():
+                    mode_impedance = {mode: purpose_impedance.pop(mode)
+                        for mode in [main_mode] + acc_modes}
+                    acc_splits[main_mode], logsum = self.split_connection_mode(
+                        mode_impedance, main_mode)
+                    purpose_impedance[main_mode] = {"logsum": logsum}
+                    mtx[main_mode] = logsum
+        return acc_splits
```
zptro: If you would make this an own method in `TourPurpose`, it could be called in both classes.

Scripts/utils/external.py
```diff
@@ -0,0 +1,59 @@
...
+import parameters.commodity as param
+
+
+def fratar(target_vect, trips, max_iter=10):
+    """Perform fratar adjustment of matrix with production target.
+
+    Parameters
+    ----------
+    target_vect : numpy/pandas array
+        Production target
```
zptro: I suggest adding the attraction target to this function already, so that this function does not have to be modified when the model is developed.

Scripts/demand/foreign_external.py
```diff
@@ -0,0 +1,85 @@
...
+        ------
+        numpy.ndarray
+            Foreign external passenger demand matrix for whole day
+        """
+        zone_data_base = self.zdata_b.get_foreign_external_data()
+        zone_data_forecast = self.zdata_f.get_foreign_external_data()
+        production_base = self._generate_trips(zone_data_base, mode)
+        production_forecast = self._generate_trips(zone_data_forecast, mode)
+
+        # NOTE: Eli tässä input-matriisissa on kaikki sentroidit, mutta nollasta poikkeavia arvoja on vain lähtömaa-sijoittelualue - ulkomaan alueklusteri -pareilla.
```
zptro: Please remove these notes.

Scripts/demand/foreign_external.py
```diff
@@ -0,0 +1,85 @@
...
+            base_mtx = numpy.array(mtx[mode]).clip(0.000001, None)
+            base_colsum = base_mtx.sum(1)
+        # Calibrate generation
+        production = calibrate(
+            base_colsum, production_base, production_forecast)
+        
+        # Get zone numbers for matrix construction
+        all_zone_numbers = numpy.array(self.zone_numbers)
+        domestic_zone_numbers = all_zone_numbers[(all_zone_numbers >= purpose_areas["domestic"][0]) & (all_zone_numbers <= purpose_areas["domestic"][1])]
+        external_zone_numbers = all_zone_numbers[(all_zone_numbers >= purpose_areas["external"][0]) & (all_zone_numbers <= purpose_areas["external"][1])]
```
zptro: Remove.

Scripts/utils/matrix_calibration.py
zptro, line 1: I realized that we still have the unused `utils.freight` module with an almost identical `fratar` function and an exactly identical `calibrate` function. Could you delete that module and maybe rename this module something more general like `matrix_calibration`?

Scripts/travel_iteration.py
```diff
@@ -129,9 +132,16 @@ def __init__(self,
...
+            if specification["name"] == "hb_abroad_other":
+                purpose = ForeignExternalPurpose(specification,
+                                                              self._zone_datas,
+                                                              self.resultdata,
+                                                              cost_data["cost_changes"],
+                                                              foreign_external_path)
+            else:
+                purpose = TravelPurpose(
+                    specification, self._zone_datas, self.resultdata,
+                    cost_data["cost_changes"])
```
zptro: I would prefer not having any logic here, but only add the new argument to `TravelPurpose`. The logic can be in  `TravelPurpose.__new__`.

zptro (dismissed):

zptro (dismissed):

zptro (approved):

PR #402, opened 2026-05-07
Refactor cost transformations

zptro:
- Move all purpose-specific parameters to json files. This finally makes it possibly to add and remove tour purposes only by adding and removing json files.
- Remove "transform" syntax from json demand model specifications and move transformations to `transform_impedance`.
- Reduce use of if and try clauses in `transform_impedance`.
- Add separate `TravelPurpose` class to avoid travel demand specifications affecting freight demand model.


Scripts/datatypes/purpose.py
```diff
@@ -156,113 +219,63 @@ def transform_impedance(self, impedance):
...
+                                               * (self.occupancy["car_drv"]-1)
+                                               / self.occupancy["car_drv"])
+            day_imp["car_pax"]["cost"] *= (self.cost_share
+                                           / self.occupancy["car_pax"])
+        for mode in day_imp:
+            if "car" in mode:
+                day_imp[mode]["time"] += numpy.asarray(zone_data["time*within_zone"])
+                day_imp[mode]["cost"] += numpy.asarray(zone_data["cost*within_zone"])
+            if mode in ("walk", "bike"):
+                day_imp[mode]["dist"] += numpy.asarray(zone_data["dist*within_zone"])
```
zptro: These should rather be added directly after parking cost and time, but that will affect results slightly.

attesn (approved):
Now this is great update. 

Do we use `SimpleTourPurpose` to anything or can we drop it?

attesn (approved):

PR #403, opened 2026-05-11
Add sec_dest_rates as empty dictionary to ExternalPurpose

zptro:


jonisylvinwsp (approved):
Let's proceed with this fix!

PR #405, opened 2026-05-19
Add higher transfer penalty from long-dist to long-dist

zptro:
This is ready to be tested in practice.


Scripts/parameters/assignment.py
```diff
@@ -217,6 +217,18 @@
...
 }
+long_dist_boarding_penalty = {
+    'b': 10, # Bus
+    'e': 30, # Coach bus
+    'g': 8, # Trunk bus
+    't': 5, # Tram
+    'p': 5, # Light rail
+    'm': 5, # Metro
+    'w': 5, # Ferry
+    'r': 5, # Commuter train
```
jalovaj2:
```suggestion
    'r': 30, # Commuter train
```

jalovaj2 (dismissed):
Looks good! The model results indicate this change has significantly reduced the transfers from Lahti train station, from roughly 5000 to 3500. Additionally, the train traffic has been reduced at Helsinki-Lahti and Lahti-Kouvola legs, as desired.

We discussed with @attesn that for long-distance trips, the commuter-commuter train transfers could also be prohibited. The model results with this modification has further reduced the transfer at Lahti train stations from 3500 to 3200. We can discuss further if this change is actually necessary, but since the train traffic around Lahti station is still overshooting, this could be reasonable

Scripts/parameters/assignment.py
zptro: The suggested parameter change does not effect commuter-commuter train transfer, but long-distance-to-commuter train transfers. I do not think this is a good idea, as it also affects transfers in the Helsinki region where I think this type of transfer is rather attractive (e.g., transferring to the commuter train going to the airport). The suggested parameter change is also not affecting the opposite direction (commuter train to long-distance transit).

jalovaj2: Thanks for your insights. Let's not update the 'r' mode transfer penalty at all.

attesn (approved):

PR #407, opened 2026-05-28
Add age-segmented size factors for visit and overnight tours

attesn:
Higher visiting tour attraction for older age groups. This factor represents behaviour where people tend to visit their family home or long-term home place.


zptro (approved):
Nice!

PR #410, opened 2026-06-09
Calc domestic leg's vehicles

Konnde:
Adds tons to vehicles calculation for foreign model's domestic leg. Because vehicle shares are different in foreign model cost specifications, don't override foreign purposes with domestic purposes.


zptro (approved):
Looks good!

PR #412, opened 2026-06-11
Use optional zonedata result aggregations if they exist

attesn:
User is able to add optional aggregations to Zonedata input with `aggregate_results` prefix and result files are then aggregated automatically, without changing the model code.


jalovaj2 (approved):
Looks good! I tested this by running it with both long-distance and short-distance models from Valma UI. For both, the input file validation works fine and the models start running normally.

PR #414, opened 2026-06-15
Update domestic freight models

Konnde:
These changes only concern model related changes and the model runs through with the introduced updates. Other features that will supplement these changes to produce refined model results will be introduced in other upcoming PRs.

This PR introduces new oljyru (öljytuotteiden runkokuljetukset - trunk routes for oil based products) model. This new model is the only model without truck mode. For this reason small refactoring has been added.

Note that this is currently in conflict with #410 in `valma_freight.py`


Scripts/utils/freight_costs.py
```diff
@@ -99,7 +99,7 @@ def calc_rail_cost(unit_costs: Dict[str, Dict],
                            + params["terminal_cost"])
     rail_cost = mode_cost["diesel_train"]
     rail_aux_cost = get_aux_cost(unit_costs, impedance, model_category)
-    return rail_cost + rail_aux_cost
+    return rail_cost, rail_aux_cost
```
zptro: I suggest returning dicts from these functions instead, then you can simplify the logic in `FreightPurpose.get_costs()`.


```suggestion
    return {"cost": rail_cost, "aux_cost": rail_aux_cost}
```

zptro (changes requested):
Just a small suggestion regarding the added logic.

Scripts/datatypes/purpose.py
```diff
@@ -772,9 +772,16 @@ def get_costs(self, impedance: dict):
...
+        costs = {mode: {"cost": calc_cost(mode, self.costdata, impedance[mode],
+                                          self.model_category)}
                 for mode in self.modes}
+        for mode in costs:
+            if isinstance(costs[mode]["cost"], dict):
+                if self.model and "aux_cost" in self.model.mode_choice_param[mode]["impedance"]:
+                    costs[mode] = costs[mode]["cost"]
+                else:
+                    costs[mode]["cost"] = sum(costs[mode]["cost"].values())
+        return costs
```
zptro: I had something like this in mind. However, we then need to change the return value of `calc_road_cost()` as well. I think that makes sense though.

```suggestion
        costs = {mode: calc_cost(mode, self.costdata, impedance[mode],
                                 self.model_category)
                for mode in self.modes}
        for mode, mode_costs in costs.items():
            if self.model and "aux_cost" not in self.model.mode_choice_param[mode]["impedance"]:
                mode_costs["cost"] += mode_costs.pop("aux_cost", 0)
        return costs
```

zptro (changes requested):

Scripts/datatypes/purpose.py
Konnde: I tried few compact alternatives similar to this but I noticed that the different requirements conflict quite a bit. In this suggestion the problem I see is that it doesn't handle a case when self.model is None (foreign models) which should output the same result when self.model and aux_cost not in choice params.

The other thing is that if we also return road costs as a dict, is that it will cascade to aux costs meaning that aux costs are then {"aux_cost": {"cost": numpy.ndarray}}. This has to be then handled. I was also thinking whether it would be better to switch this handling to domestic model's calc_traffic method. Then self.model would always be true but then again it would be nicer to have cost related handling in this cost getter method.

zptro: Good points!

As for the first point, I had messed up the logic. My if-clause should be `if not self.model or "aux_cost" not in self.model.mode_choice_param[mode]["impedance"]`. Would that solve the problem?

The second point is more intricate. One solution would be to retain `calc_road_cost()` as is, but change one line in `calc_cost()`: `return {"cost": calc_road_cost(unit_costs, impedance, model_category)}`.

Konnde: It does. That's actually very smart; I hadn't thought about combination of summing and pop with 0 fallback when, for example, truck doesn't have aux cost. I'll push a new suggestion.

Scripts/utils/freight_costs.py
```diff
@@ -25,16 +25,18 @@ def calc_cost(mode: str, unit_costs: Dict[str, Dict],
...
+            return {"cost": calc_road_cost(unit_costs, impedance, model_category)}
         case "freight_train":
-            return calc_rail_cost(unit_costs, impedance, model_category)
+            rail_cost, aux_cost = calc_rail_cost(unit_costs, impedance, model_category)
+            return {"cost": rail_cost, "aux_cost": aux_cost}
         case "ship":
-            return get_domestic_ship_cost(unit_costs, impedance, model_category)
+            ship_cost, aux_cost = get_domestic_ship_cost(unit_costs, impedance, model_category)
+            return {"cost": ship_cost, "aux_cost": aux_cost}
         case _:
```
Konnde, line 38: How about this structure. Retain common type returning from each mode specific cost calc function (numpy mtx) and transform each one here into dict. Would keep the structure same for each mode both here and in each cost calc function.

zptro: That looks nice! 👍

zptro (approved):

PR #417, opened 2026-06-23
Refactor handling of extra attributes and network fields

zptro:
- Move all assignment results to network fields
   - Reduce needs to reserve extra-attribute space in advance
   - Enable longer assignment-class names @attesn 
- Remove `AssignmentPeriod.transit_results_links_nodes()` and `AssignmentPeriod._calc_transit_link_results()` -> Integrate both in `TransitMode.calc_transit_network_results()` -> Get assignment-class-specific aux volumes as bonus


Scripts/assignment/freight_assignment.py
```diff
@@ -61,16 +61,33 @@ def save_network_volumes(self, commodity_class: str):
...
-            spec["on_segments"]["transit_volumes"] = "@" + attr_name
-            spec["on_links"]["aux_transit_volumes"] = "@a_" + attr_name
             self.emme_project.network_results(
                 spec, self.emme_scenario, ass_class)
-        attr_name = (commodity_class + "truck")[:17]
+            seg_attr = "#" + commodity_class + ass_class
+            self.emme_project.create_network_field(
+                    "TRANSIT_SEGMENT", "REAL", seg_attr, "commodity flow",
+                    overwrite=True, scenario=self.emme_scenario)
+            link_attr = "#aux_" + commodity_class + ass_class
```
Konnde: I think adding underscore to seg_attr and link_attr between commodity class and assignment class would be nice visual assistance for differentiating between large amount of network fields within emme tables.

#elektr_truck rather than #elektrtruck

Konnde (dismissed):
Everything seems good here. I also did few assignment tests and didn't encounter any problems here nor in emme (bank full etc). Very nice updates. Just one suggestion on my part.

Scripts/assignment/freight_assignment.py
zptro: Good point! Fixed.

Konnde (approved):

PR #418, opened 2026-06-24
Force transit mode boarding

Konnde:
Force freight spec transit mode boarding before being able to reach destination.


Scripts/parameters/assignment.py
```diff
@@ -447,6 +447,7 @@
     'a'
 ]
 park_and_ride_mode = 'u'
+freight_aux_mode = "F"
```
Konnde: This "F" mode is general freight aux mode. However timber specific (commodity metsat) specific "T" mode also exists on the network. Their difference is that T-mode has freight loading/unloading stations that are operating only for timber. I however am unsure how to make this change so that metsat-specific freight_train and ship impedances are returned.

Here is a simple suggestion how these could the modes could be described here
```
freight_aux_modes = {
    "general": "F",
    "metsat": "T"
}
```

Scripts/assignment/datatypes/freight_specification.py
zptro, line 48: We do not have unique copy of this for each journey level, so I think changing a transition rule will change it for all levels.

Scripts/assignment/datatypes/freight_specification.py
```diff
@@ -88,7 +92,7 @@ def __init__(self, name: str, assignment_period: AssignmentPeriod,
                 "perception_factor": 1,
             },
             "aux_transit_by_mode": [{
-                "mode": param.park_and_ride_mode,
+                "mode": param.freight_aux_mode,
```
zptro: `park_and_ride_mode` is set according to mode `k` here: https://github.com/Traficom/valma-model-system/blob/dev/Scripts/assignment/freight_assignment.py#L119. We should change that if we want to start using this new mode.

Konnde: Okay this is good to know. The testing we have done so far has shown that currently there is little difference in results between the two modes. In that sense we could postpone that change to be done after holidays. Or would it be wiser to do the implementation here. What do you think @zptro?

zptro: How has F mode been defined in the network you are using?

Konnde: We're using koko suomi network, and I actually don't know the detailed conditions on how network builder adds these modes. Should ask Miikka and Julius.

PR #419, opened 2026-06-24
New mode specific intrazonal distances

attesn:
First step to correct the subarea mode shares. This fixes the estimation for visit tours and reduces other possobile effects of unrealistic walking distances to estimations.

Makes calculation of intrazonal distances more difficult to external users, so we should examine if we move to using zone radius (`sqrt land area`) and separate coefficient later. 

New `zonedata.gpkg` files are at Sharefile with tag "--intrafix".


zptro (approved):
Looks good!

PR #422, opened 2026-06-25
Update travel demand model parameters

attesn:
Several changes to demand model parameters:
* Mode constants to oversampled areas that have distinct mode shares
* Remove population from grocery size factor.
* Change student size variables from number of workers to students. 
* Density attraction variable to bike utility.
* Rename overnight purpose to better match the tour building.
* Use impedance matrices from #405


zptro (dismissed):

Scripts/datahandling/zone_variables.json
```diff
@@ -52,12 +51,14 @@
             "intra_dist_bike",
             "intra_dist_car",
             "sh_office",
-            "sh_row_or_detached"
+            "sh_row_or_detached",
+            "density_pop_wrk"
```
zptro: I would prefer having population and workplace density calculated in the `zonedata` module, because otherwise it will easily become inconsistent with user changes of population and workplaces.

zptro (approved):
🚀

PR #423, opened 2026-06-29
Refactor/freight model

zptro:
- Change name of `FreightPurpose` to `FreightCommodity` and move it to own module `commodity`
- Split into two sub-classes `DomesticCommodity` and `ForeignCommodity`
- Make `write_trade_route_summary` and `write_purpose_summary` class methods
- Call `run_logistics_module` from inside `calc_traffic`
- Call `write_trade_route_summary` from inside `run_trade_route_module`


Scripts/valma_freight.py
Konnde: Drop deprecated import calls for `write_purpose_summary`, `write_zone_summary` and `clusters`

Konnde (changes requested):
Very nice refactoring! I was also at some point thinking if it would be wise to restructure FreightPurpose little bit as it was mixing both domestic and foreign method calls. Good to see that you had similar thoughts. Understanding which methods belong to which freight "model type" would be somewhat confusing for people not accustomed with the development here, so glad to see that this branch fixes that. 

The restructuring looks good here but do fix the import error in `valma_freight.py`. Could also remove few unnecessary imports from `freight_utils.py` while at it.

Konnde (approved):
Looks good!

PR #424, opened 2026-06-30
Integrate freight cost utils in object-oriented framework

zptro:
- Move freight cost calculations into `FreightCommodity`. Removes the need to send already-existing object variables as arguments to sub-functions.
- Generalize the train and ship calculations in one method `_calc_ton_cost`.
- Move "aux_cost" filter logic to assignment side, as it is not commodity-specific. This causes some changes in test results, as I did not update the impedance matrices accordingly.
- Remove some of the dependencies from foreign ship cost calculation to other calculations.
- Remove variable `model_category`. Makes the differences between `DomesticCommodity` and `ForeignCommodity` more clear.

Notes
- In fact, I think `_calc_foreign_ship_cost` should be moved to the assignment side, as the unit costs are identical for all commodities (except the harbor fee, which is common for all ship types). This will make draught handling even more straightforward. This could be done in a separate PR.
- The different handling of "terminal_cost" causes a small change in the results. I have not figured out why this happens. It might be only a rounding issue.
- Further notes below.

No hurry with reviewing this one!


Scripts/assignment/datatypes/freight_specification.py
```diff
@@ -142,12 +143,15 @@ def __init__(self, name: str, assignment_period: AssignmentPeriod,
...
             **self.time.item,
             **self.canal_cost.item,
         }
+        dist = self.dist.data
+        aux_dist = self.aux_dist.data
         aux_cost = (self._time_unit_cost*self.aux_time.data/60
-                    + self._dist_unit_cost*self.aux_dist.data)
+                    + self._dist_unit_cost*aux_dist)
+        aux_cost = numpy.where(
+            (aux_dist > dist*2) | (dist == 0), numpy.inf, aux_cost)
```
zptro, line 154: The `dist == 0` condition can be removed in #427

Scripts/datatypes/commodity.py
```diff
@@ -147,13 +120,96 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
+            costs[truck] = {"cost": self._transform_road_cost(
+                impedance[truck]["cost"])}
+        train = "freight_train"
+        if train in self.modes:
+            costs[train] = self._calc_ton_cost(
+                impedance[train], self.costdata[train]["diesel_train"])
+        ship = "ship"
+        if ship in self.modes:
+            costs[ship] = self._calc_ton_cost(
+                impedance[ship], self.costdata[ship]["domestic_vessel"])
```
zptro, line 176: After merging of #427, these (and "timber_train") become identical, so then they can simply be looped.

Scripts/datatypes/commodity.py
```diff
@@ -147,13 +120,96 @@ def calc_vehicles(self, matrix: numpy.ndarray, ass_class: str):
...
+        unit_costs : Dict[str, float]
+            Impedance type (time/dist/toll_cost) : unit cost value
+
+        Returns
+        ----------
+        Dict[str, numpy.ndarray]
+            auxiliary road cost : numpy.ndarray
+            impedance type cost : numpy.ndarray
+        """
+        impedance["terminal_cost"] = numpy.ones_like(impedance["dist"])
```
zptro, line 204: Temporal solution. After #427, terminal visit numbers will come from assignment.

Scripts/assignment/datatypes/freight_specification.py
```diff
@@ -142,12 +143,15 @@ def __init__(self, name: str, assignment_period: AssignmentPeriod,
     def get_matrices(self):
         mtxs = {
             **self.dist.item,
-            **self.aux_dist.item,
```
Konnde: Removing this actually now causes KeyError here during summary writing. https://github.com/Traficom/valma-model-system/blob/4a1d33fecbd1839f732291c9f936bfb2a4813c0b/Scripts/datatypes/commodity.py#L297

Konnde (changes requested):
Went through the cost changes and they seem ok. Basically the focus is shifting from few different types of feature files to an expansive commodity feature. Few parts are referencing upcoming feature #427 so I will do more throughout testing when reviewing that one.

There is one KeyError which needs to be handled before merging.

Scripts/assignment/datatypes/freight_specification.py
zptro: Thanks for pointing this out! Now fixed.

Konnde (approved):
Looks good!

PR #427, opened 2026-07-03
Improve freight route choice

zptro:
- Get separate matrices for diesel train and electric train impedance, depending on the distance traveled with each mode in assignment. This also includes the number of terminal visits per train type. Their costs can then be summed per commodity.
- Make destinations unreachable from journey level 0 (truck access). Add journey level for truck egress.
- Make route choices consistent with demand model by adding distance cost to disutility. For ships, this total cost could potentially be used directly in demand model. For trains, the situation is a bit more complicated, because they have commodity-specific wagon costs.
- Add separate timber-train assignment class. After #424 is merged, this can be used in demand model.
- Add terminal mode for accessing terminals. This is separate for timber train.


Scripts/valma_freight.py
```diff
@@ -87,7 +87,13 @@ def main(args):
             store_demand.store(mode, demand[mode], omx_filename, commodity.name)
         if commodity.name in args.specify_commodity_names:
             ass_model.freight_network.save_network_volumes(commodity.name)
-        
+        for mode in impedance:
+            dist = 0
+            for imp_type in impedance[mode]:
+                if "dist" in imp_type:
+                    dist += impedance[mode][imp_type]
+            if dist is not 0:
```
Konnde: Python announces SyntaxWarning for this as 0 is a literal and shouldn't be compared with is. This warning is announced in launch as well as during debugging. While this works I think we should fix the typing here, maybe something like:

`if isinstance(dist, numpy.ndarray) and numpy.any(dist):`

Konnde (commented):
I feel that I am now done with the review here. There are a lot structural changes happening in this PR and they're mostly assignment related which I am still not too knowledgeable. Based on my interpretation general complexity increases but at the same time many changes are mechanically required to enable model functionalities that are necessary for our real world assessments. 

I mentioned that the results here were different from our separate dummy model which we have been using as a platform for this. However the differences were due to me using incorrect (too cheap) truck unit costs. With correct truck unit costs the run shows similar results with dummy model's results. Based on initial results, ship unit costs seem too expensive but this is something we can deal outside this PR.

As a conclusion I think this structural update answers to our needs to force train and ship modes to use their own infrastructure. Also enabling separation of freight train and timber train is an important addition here.

I added a note on python syntax warning which can be updated if you see fit.

Konnde (approved):

Konnde (approved):
Looks good!

PR #428, opened 2026-08-05
Add sqrt density to car ownership model

jonisylvinwsp:
-Replace accessibility in car ownership model with sqrt pop density
-Update car ownership model parameters and calibrate it
-Update mode share for tests


attesn (approved):
Looks good! Validation to Statfi data was similar as before. Carless households are prediced correctly, but some differences between model and validation data in 1 and 2+ households.

PR #432, opened 2026-08-10
Omit accessibility export for external travel groups

jalovaj2:
Fixes a fatal error when trying to export accessibility values for external travel groups (hb_abroad_other) with no destination choice model.


zptro (changes requested):
I would prefer fixing this by initializing `ModeDestModel.accessibility` as an empty dict in `__init__()` method instead of a hard-coded if-clause.

zptro (approved):
Great, thanks!

PR #433, opened 2026-08-11
Refactor logit utility calculation

zptro:
After removing legacy Helmet functionality (#192), the structure of utility calculations can be simplified substantially.


attesn (approved):
Good update!

PR #436, opened 2026-08-12
Set car driver unavailable for under 18 years old

attesn:
Restrict availability of car driver mode in estimation.  This makes estimation results considerably better and can have some impact on underestimating car use in rural areas (cannot be validated though).  However, this model-system solution does not work correctly, as the share should not be share of 7-17 population, but share of purpose tours made by 7-17 year old population.

Not sure if we are able to get working solution without agent based model or adding too much complexity.


zptro (dismissed):
This seems like an acceptable solution. You also seem to have formed the mutually exclusive segments correctly.

attesn:
This did not fix what it was supposed to, so declining request. We'll return on this subject when agent model is developed.

PR #444, opened 2026-08-24
Refine parking cost logic for park-and-ride

jalovaj2:
This PR fixes multiple issues with existing park-and-ride parking costs that affect the access mode shares for park-and-ride.

1. Set park-and-ride parking cost (from #park_cost_n) only for park-and-ride transfers. Previously park-and-ride parking costs where applied for transfers between all modes.

2. Set park-and-ride parking cost (from #park_cost_n) only for park-and-ride transfers to ship (mode "d") where the people park their cars at the harbor (that is, change mode from 'u' to 'd' at the harbor). Previously transfer parking costs where applied also for preceding transit mode transfers and for cars boarded to ships

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -353,10 +353,21 @@ class AssignmentPeriod(Period):
             network.mode(param.assignment_modes["car"]),
             network.mode(param.assignment_modes["truck"])
         }
+        long_dist_terminal_modes = {network.mode(mode_id) for mode_id in param.long_dist_terminal_modes}
         park_and_ride_mode = network.mode(param.park_and_ride_mode)
+        long_distance_nodes = {node.id for node in network.nodes()
+                                if any(segment.line.mode in long_dist_terminal_modes
+                                       for link in node.incoming_links()
+                                       for segment in link.segments())}
         car_time_zero = []
         for link in network.links():
             linktype = link.type % 100
+            this_link_modes = {segment.line.mode for segment in link.segments()}
+            is_transit_line_link = this_link_modes & long_dist_terminal_modes
+            connects_long_distance = (
+                link.i_node.id in long_distance_nodes
+                or link.j_node.id in long_distance_nodes)
+            is_park_and_ride_connector = connects_long_distance and not is_transit_line_link
             if link.type > 80 and linktype in param.roadclasses:
                 # Car link with standard attributes
                 roadclass = param.roadclasses[linktype]
@@ -394,6 +405,9 @@ class AssignmentPeriod(Period):
             else:
                 # Link with no car traffic
                 link.volume_delay_func = 0
+            if is_park_and_ride_connector:
+                # Park-and-ride connectors
+                link.volume_delay_func = 97
             if link["#buslane"]:
                 if (link.num_lanes == 3
                         and roadclass.num_lanes == ">=3"):
@@ -412,7 +426,7 @@ class AssignmentPeriod(Period):
                         msg = f"Car travel time on link {link.id} is {car_time}"
                         log.error(msg)
                         raise ValueError(msg)
-            if car_modes & link.modes:
+            if (car_modes & link.modes or connects_long_distance) and not is_transit_line_link:
                 link.modes |= {main_mode, park_and_ride_mode}
             else:
                 link.modes -= {main_mode, park_and_ride_mode}
--- a/Scripts/assignment/datatypes/journey_level.py
+++ b/Scripts/assignment/datatypes/journey_level.py
@@ -83,20 +83,19 @@ class JourneyLevel:
             } for mode in param.long_dist_transit_modes[transit_class]]
         if park_and_ride:
             if transit_class in param.car_access_classes:
-                # Park-and-ride (car) mode allowed only on level 0.
-                car = FORBIDDEN if level >= PARKED else NOT_BOARDED
+                car = FORBIDDEN if level >= PARKED else PARKED
                 # If we want parking to be allowed only on specific links
                 # (i.e., park-and-ride facilities), we should specify an
                 # own mode for these links. For now, parking is allowed
                 # on all links where walking to a stop is possible.
-                walk = PARKED if level == NOT_BOARDED else level
+                walk = FORBIDDEN if level == PARKED else level
             elif transit_class in param.car_egress_classes:
                 # Transfer to park-and-ride (car) mode only allowed after first
                 # boarding. If we want parking to be allowed only on specific
                 # links, we should specify an own mode for these links.
                 # For now, parking is allowed on all links where walking
                 # from a stop is possible.
-                car = LEFT if BOARDED_LOCAL < level < FORBIDDEN else FORBIDDEN
+                car = LEFT if BOARDED_LOCAL <= level < FORBIDDEN else FORBIDDEN # A bug fix to fix assymmetry between access and egress? With this fix both now allow BOARDED_LOCAL where as before only access allowed that.
                 walk = FORBIDDEN if level == LEFT else level
             transitions.append({
                 "mode": param.park_and_ride_mode,
--- a/Scripts/parameters/assignment.py
+++ b/Scripts/parameters/assignment.py
@@ -98,6 +98,7 @@ volume_delay_funcs = {
     "fd11": "length*(60/ul2)+el1",
     "fd90": "length*(60/ul2)",
     "fd91": "length*(60/ul2)",
+    "fd97": "length*(60/5)",
     "fd99": "length*(60/ul2)",
     # Bike functions
     "fd70": "length*(60/19)",
@@ -455,6 +456,9 @@ long_dist_transit_modes = {
     "pt_car_egr": ['j'],
     "pt_taxi_egr": ['e', 'j'],
     "airpl_car_egr": ['l'],
+}
+long_dist_terminal_modes = {
+    'l', 'j', 'd'
 }
 aux_modes = [
     'a'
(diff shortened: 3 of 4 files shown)


jalovaj2:
It seems that currently no park-and-ride transfers occur in ship terminals. This could be due to relatively high parking costs that were not included in the estimation, or could be a bug with the implementation. However, the park-and-ride transfers at airports seem normal.

Scripts/assignment/datatypes/journey_level.py
```diff
@@ -83,20 +83,19 @@ def __new__(self, level: int, transit_class: str,
...
             if transit_class in param.car_access_classes:
-                # Park-and-ride (car) mode allowed only on level 0.
-                car = FORBIDDEN if level >= PARKED else NOT_BOARDED
+                car = FORBIDDEN if level >= PARKED else PARKED
                 # If we want parking to be allowed only on specific links
                 # (i.e., park-and-ride facilities), we should specify an
                 # own mode for these links. For now, parking is allowed
                 # on all links where walking to a stop is possible.
-                walk = PARKED if level == NOT_BOARDED else level
+                walk = FORBIDDEN if level == PARKED else level
```
zptro: I think this should be `walk = FORBIDDEN if level < PARKED else level`, because we want to make sure people cannot avoid the parking cost by walking past them.

Scripts/assignment/datatypes/journey_level.py
```diff
@@ -83,20 +83,19 @@ def __new__(self, level: int, transit_class: str,
             } for mode in param.long_dist_transit_modes[transit_class]]
         if park_and_ride:
             if transit_class in param.car_access_classes:
-                # Park-and-ride (car) mode allowed only on level 0.
-                car = FORBIDDEN if level >= PARKED else NOT_BOARDED
+                car = FORBIDDEN if level >= PARKED else PARKED
```
zptro: This seems like an unnecessary and faulty change. I think this implies that you will not be able to use car as access mode, because if you try to use it, you will immediately be moved to a level where it is forbidden.

Scripts/assignment/datatypes/journey_level.py
```diff
@@ -83,20 +83,19 @@ def __new__(self, level: int, transit_class: str,
...
-                walk = PARKED if level == NOT_BOARDED else level
+                walk = FORBIDDEN if level == PARKED else level
             elif transit_class in param.car_egress_classes:
                 # Transfer to park-and-ride (car) mode only allowed after first
                 # boarding. If we want parking to be allowed only on specific
                 # links, we should specify an own mode for these links.
                 # For now, parking is allowed on all links where walking
                 # from a stop is possible.
-                car = LEFT if BOARDED_LOCAL < level < FORBIDDEN else FORBIDDEN
+                car = LEFT if BOARDED_LOCAL <= level < FORBIDDEN else FORBIDDEN # A bug fix to fix assymmetry between access and egress? With this fix both now allow BOARDED_LOCAL where as before only access allowed that.
```
zptro: `BOARDED_LOCAL` is the level that comes before `BOARDED_LONG_D`. We do not want car egress from that level. Please revert.

Scripts/assignment/datatypes/journey_level.py
```diff
@@ -83,20 +83,19 @@ def __new__(self, level: int, transit_class: str,
...
                 walk = FORBIDDEN if level == LEFT else level
```
zptro: To get this symmetric with access, I think we need to change it like this:


```suggestion
                walk = FORBIDDEN if level >= BOARDED_LONG_D else level
```

Scripts/assignment/assignment_period.py
```diff
@@ -353,10 +353,21 @@ def _set_car_vdfs(self, use_free_flow_speeds: bool = False):
             network.mode(param.assignment_modes["car"]),
             network.mode(param.assignment_modes["truck"])
         }
+        long_dist_terminal_modes = {network.mode(mode_id) for mode_id in param.long_dist_terminal_modes}
         park_and_ride_mode = network.mode(param.park_and_ride_mode)
+        long_distance_nodes = {node.id for node in network.nodes()
+                                if any(segment.line.mode in long_dist_terminal_modes
+                                       for link in node.incoming_links()
+                                       for segment in link.segments())}
```
zptro: This seems slow. Could we use link modes instead?

Scripts/assignment/assignment_period.py
```diff
@@ -412,7 +426,7 @@ def _set_car_vdfs(self, use_free_flow_speeds: bool = False):
                         msg = f"Car travel time on link {link.id} is {car_time}"
                         log.error(msg)
                         raise ValueError(msg)
-            if car_modes & link.modes:
+            if (car_modes & link.modes or connects_long_distance) and not is_transit_line_link:
```
zptro: Does this imply that car mode will be prohibited on ferry links? I thought we wanted to allow them there?

Scripts/parameters/assignment.py
```diff
@@ -456,6 +457,9 @@
     "pt_taxi_egr": ['e', 'j'],
```
zptro: I realize I have forgot to add 'd' to these. That probably explains why you get no park-and-ride in the harbors.

zptro (changes requested):
The journey-level approach is not easy to comprehend! Here are some comments.

zptro:
> 1. Set park-and-ride parking cost (from #park_cost_n) only for park-and-ride transfers. Previously park-and-ride parking costs where applied for transfers between all modes.

You mean that you set it for car mode, but not for walk mode, right?

> 2. Set park-and-ride parking cost (from #park_cost_n) only for park-and-ride transfers to ship (mode "d"). Previously transfer parking costs where applied also for cars boarded to ships.

I do not think this is what we want to achieve. We want the parking fee to apply to people parking in the harbor, but not to people parking at a train station elsewhere in Finland and _taking the tram from the train station_.

Scripts/assignment/assignment_period.py
jalovaj2: If you mean car mode as param.assignment_modes["car"] then I think no because this part of the code only adds (or removes) param.park_and_ride_mode from the link, not param.assignment_modes["car"].

If you mean car mode as param.park_and_ride_mode then yes, 'u' mode will not be allowed on links that have ferry lines (is_transit_line_link).

The idea (together with journey level modifications) is to allow 'u' mode on the link that is used to access the node where the ferry line starts, and then after that 'u' mode will not be available on the link that has ferry link. This way the park-and-ride trip chain uses 'u' mode until accessing the node where the transit line starts and the parking costs apply only for the 'u' mode. Transfer to 'a' mode will be prohibited when using 'u' so that you always get the parking cost if you use 'u' mode to get to a transit line that has parking cost involved (i.e. use 'u' mode to get to a node with parking cost where you switch to long distance transit line ).

This is also the reason why the vdf for the links used to access terminal nodes is changed to match the walking speed so that they function as a connectors to terminals when using 'u' or 'a' to access the terminal

zptro: Sorry for the ambiguity. I was referring to `main_mode` ("h"), which have to be allowed in car assignment.

Scripts/assignment/assignment_period.py
jalovaj2: I noticed that having the mode allowed on a link doesn't guarantee there is a transit line with that allowed mode on that link. I nevertheless optimized it to be significantly faster than the previous version

Scripts/assignment/assignment_period.py
jalovaj2: Good point! I will add main_mode ("h") to ferry links as intended

Scripts/assignment/datatypes/journey_level.py
```diff
@@ -83,20 +83,19 @@ def __new__(self, level: int, transit_class: str,
...
             if transit_class in param.car_access_classes:
-                # Park-and-ride (car) mode allowed only on level 0.
-                car = FORBIDDEN if level >= PARKED else NOT_BOARDED
+                car = FORBIDDEN if level > ENTERED_CAR else ENTERED_CAR
                 # If we want parking to be allowed only on specific links
                 # (i.e., park-and-ride facilities), we should specify an
                 # own mode for these links. For now, parking is allowed
                 # on all links where walking to a stop is possible.
-                walk = PARKED if level == NOT_BOARDED else level
+                walk = FORBIDDEN if level == ENTERED_CAR else level
```
zptro: I think it would be better to simply remove journey level 1 altogether. In the current solution the parking costs can be avoided by staying on journey level 0 (walking) until boarding.

zptro (changes requested):

Scripts/assignment/datatypes/journey_level.py
jalovaj2: That is true. Let's remove it

Scripts/assignment/assignment_period.py
```diff
@@ -413,6 +430,10 @@ def _set_car_vdfs(self, use_free_flow_speeds: bool = False):
                         log.error(msg)
                         raise ValueError(msg)
             if car_modes & link.modes:
+                link.modes |= {main_mode}
+                if not is_transit_line_link:
+                    link.modes |= {park_and_ride_mode}
+            elif connects_long_distance and not is_transit_line_link:
```
zptro:
```suggestion
            elif is_park_and_ride_connector:
```

Scripts/assignment/datatypes/journey_level.py
```diff
@@ -5,30 +5,27 @@
...
-    BOARDED_LONG_D: True,
-    BOARDED_DEST: True,
-    LEFT: True,
-    FORBIDDEN: False,
+    NOT_BOARDED: False, # 0
+    BOARDED_LOCAL: True, # 1
+    BOARDED_LONG_D: True, # 2
+    BOARDED_DEST: True, # 3
+    LEFT: True, # 4
+    FORBIDDEN: False, # 5
```
zptro: Really small thing, but could you move the number comments to the first tuple (rows 8-12)

zptro (approved):
Excellent!

